PR #7, opened 2017-08-18
Move python package installation to requirements.txt

dctrud:
To allow dev work in non-docker environment, it's nice to have the requirements installed from a requirements.txt instead of directly with RUN pip install.

**Also** - changes cron install to anacron in intial apt-get so not so many layers to rebuild on code change.


dctrud:
Oh, while I rembember... this pins celery at 3.x (latest 4.x will give import error from django-celery). Also grabs your development singularity-python which has the singularity.registry stuff

Dockerfile
```diff
@@ -13,77 +14,29 @@ RUN apt-get update && apt-get install -y \
...
-RUN pip install python3-saml
-
 # Install Singularity
 RUN git clone https://www.github.com/singularityware/singularity.git
 WORKDIR singularity
 RUN ./autogen.sh && ./configure --prefix=/usr/local && make && make install
 
+# Install Python requirements out of /tmp so not triggered if other contents of /code change
+ADD requirements.txt /tmp/requirements.txt
+RUN pip install -r /tmp/requirements.txt
```
vsoch: This is a good idea! Could we add the requirements.txt to /code instead, so the versions are preserved with the application?

dctrud: The `ADD . /code` further down the Dockerfile will bring requirements.txt into /code later on. I bring into /tmp here earlier so that the requirements layer is created early - and is not going to have to be rebuilt as often.

dctrud: It's one of the docker best-practices where they give an example.... but actually they prefer COPY to ADD also.

https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy

vsoch: Haha, I'm not even that picky. I'm convinced, this looks good!

PR #39, opened 2017-10-10
Dockerfile adds gsettings.m4, fixes build error

smoe:
The only idea coming to mind why on this end the build fails but not for @vsoch and friends are differences between the dynamically assigned Debian mirrors that depend on where one is located. That line fixed it (except that I had it on a separate RUN apt-get ... invocation).


vsoch:
It seems reproducibility is not location agnostic! This looks good, I'll merge and we can check if the Docker Hub build is still ok. Thanks for figuring this out!

Issue #14, opened 2017-08-22
Proposal - Container security scanning with Clair


dctrud:
[Clair](https://github.com/coreos/clair) is the CoreOS project for security static analysis of containers, scanning them for security issues (from databases of known CVEs). I'd like to propose adding support to sregistry for scanning containers using Clair.

Though Clair is centered around docker or appc images, it has been used to scan openvz templates, which are .tar archives - see [FastVPSEestiOu/check_openvz_mirror_with_clair](https://github.com/FastVPSEestiOu/check_openvz_mirror_with_clair). I'm pretty sure something similar could be done for singularity images.

This is something I'm planning to work on, and thought I'd add a ticket here in case it's of interest to others / there are any thoughts? I'm thinking I will be working to:

- [ ] Create a stand alone python tool to scan a singularity image via Clair API
- [ ] Investigating contributing this to singularity-python (if it's of interest there)
- [ ] Integrate with sregistry - celery jobs for Clair scanning periodically / on push?

Would welcome any input on if this is of interest for sregistry, or more generally.

vsoch:
I've tried out Clair (I made a [Singularity version](https://github.com/vsoch/singularity-clair) that was going to be for a check) and I stopped because it was requiring a server. This would definitely be great, but a few thoughts:

- I'm not sure that the registry server itself is the place to have this, although it could be deployed alongside a registry. We would arguably want a check to happen before it is uploaded to the registry, and the entire purpose of clair is to provide a server so the (user) could do this.
 - If it is most logical to integrate with the registry, I think it should be optional, but not default. Eg, there is another folder with a Docker-compose file to deploy it, and the docker-compose / settings should have an environment variable that turns checks on/off. 

My general thinking is that I like the idea, but I think it should be a separate module from the registry itself. Eg, it really should be running on its own server, and then have different (independent) commands run:

```
sregistry check continer.img clair
sregistry push continer.img --name library/collection:tag
```
The command line client (sregistry) would have two different servers / applications it is pinging to get that functionality. This also makes sense so we don't develop a hairball of a mess of different applications under one roof. The registry should accept and serve images, and be optimized for that. The security bits (+1 for important!) should be done separate to that.

Your approach sounds good! I can definitely add the endpoint for checks to the sregistry client. To be clear on your last point, I don't think the right place is with the singularity registry application itself, but for the client (sregistry).

[...]

PR #49, opened 2017-10-19
[WIP] LDAP Authentication & Plugin framework

dctrud:
This is work in progress for discussion.

PR provides a way of logging in to sregistry with authentication against an LDAP directory.

image

image

Working for auth against our OpenLDAP based directory. I can login nicely with my HPC LDAP account.

**Things to do / for discussion**

- I've tried here to add minimal code, so it's tied up with the .py files used for social auth. Is it preferable to remove all LDAP things into a more separated namespace?

- Doesn't yet work nicely with the agreement to terms page. I'm not sure about this though - we probably don't want an agreement to terms step here at all. Anyone who signs in via LDAP has an institutional account, and agreed to unversity-wide policy for use of systems when they got that.

- I added a `requirements_ldap.txt` with the requirements for LDAP auth - but I'm guessing these could be added into requirements.txt? Since this is dockerized there's not much harm in building with LDAP support by default?

- I added a `secrets.py.ldap` with example template LDAP configuration. Do you want to keep this or only have in docs? Where do you want the docs to go? Separate LDAP page?


vsoch:
The key is great!

In terms of organization, here is my thinking:
- if we are adding dependencies for the entire application, we should specify the exact version that we know to be working. With smaller packages like this I've seen them act unpredictably and we don't want a future build to break.
- across the site, if ldap code is mixed with standard pages (eg the login screen) there should be one boolean trigger that determines if it's rendered or not (I think you've already done this mostly)
- it's important, to keep setup simple for the user, to maintain one config.py file. For "Secrets" you would just write the variable name and set it to None. Even if it's long, if we have very clear headers for sections and then notes like "You don't need to configure this section if `ENABLE_LDAP=False`. Maybe a section would look like:

```
########################################################################
# LDAP AUTHENTICATION
ENABLE_LDAP=False
########################################################################
# LDAP authentication is... you don't need to configure this section if ENABLE_LDAP is False

```
- for the additional pages, urls, I think we will do a cleaner job if each plugin is packaged in its own namespace. Then when the server starts up, that same boolean should determine whether or not to add the module as an application. We probably want to avoid sregistry turning into a hairball of things that aren't wanted by everyone, each with a custom way to add/remove. For organization, maybe something like this:

```
shub/
├── apps
├── plugins
├── celery.py
├── __init__.py
├── logger.py
├── settings
├── urls.py
└── wsgi.py
```
And it's understood that a plugin is an optional thing, and apps tend to be (mostly) core, with some if statements in templates.

Then for the urls.py, we use the same convention - we check the global boolean and add to the dispatcher if it's True.

>> I've tried here to add minimal code, so it's tied up with the .py files used for social auth. Is it preferable to remove all LDAP things into a more separated namespace?

It would be great if ldap could be under `plugins/ldap` given that it has its own urls and templates. It's also nice to be able to find things when you need to!

>> Doesn't yet work nicely with the agreement to terms page. I'm not sure about this though - we probably don't want an agreement to terms step here at all. Anyone who signs in via LDAP has an institutional account, and agreed to unversity-wide policy for use of systems when they got that.

I think the LDAP module should have some agreement page that shows the same thing - it's more about reminding the user to to not use their institution's resource maliciously, etc.

>> I added a requirements_ldap.txt with the requirements for LDAP auth - but I'm guessing these could be added into requirements.txt? Since this is dockerized there's not much harm in building with LDAP support by default?

This is what I was referencing about with regard to adding dependencies. Another idea to consider would be to have a simple (config) text file (akin to Singularity) that gets parsed on some "install" step and dumps these variables into the application. It would maybe make sense for plugins? Then when the container is built, it references that.

>> I added a secrets.py.ldap with example template LDAP configuration. Do you want to keep this or only have in docs? Where do you want the docs to go? Separate LDAP page?

Let's make a Plugins section under docs/README.md, and then a separate link for each.

This is so great! Let me know your thoughts to the above.

vsoch:
One more quick thought! By not using python social auth, the functions that are mixed around the site that require checking the user's account would break (and we will find this during testing). I wonder if it's possible to wrap the code in a custom Auth backend? This would also fix the issue with agreeing to the terms, and also verify email checking. One of the steps I added to the pipeline was to make sure that if a user registered with two different backends but with the same email, it was associated with the same user. So either an LDAP login must ONLY be that, or integration with the social auth pipeline. What we don't want happening is a user is able to log in with LDAP, and then (some other social auth) and have two accounts created.

dctrud:
I'll work on separating it out into a plugin structure. To be honest I'm not convinced LDAP auth needs to be a plugin (I personally think LDAP is more a core feature than social auth for private registries) - but that's an unimportant difference of opinion, and having plugins is definitely important :-)

> One more quick thought! By not using python social auth, the functions that are mixed around the site that require checking the user's account would break (and we will find this during testing). I wonder if it's possible to wrap the code in a custom Auth backend?

We could certainly write a custom social-auth backend for LDAP - but that's then really recreating the stuff we get from standard django-auth-ldap and the django.contrib.auth login view (or wrapping them in spaghetti). I don't like reinventing that stuff as LDAP auth on django is really a solved problem with a widely used solution that's nicely maintained. I think it'd be nicer if any checks work with arbitrary AUTHENTICATION_BACKENDS.

I'm not clear why functions checking a user account would break? I think social-auth is integrating with the standard Django accounts models?. Do you have any example I can take a look at and think about?

vsoch:
I am using a custom user model, and django social auth with pipelines, you can read more about here --> https://python-social-auth.readthedocs.io/en/latest/pipeline.html#extending-the-pipeline.

I think for now it would be best to implement as modular, and then write up a clean doc that describes how to get it up and running, then I can test and see how it's integrating with a standard user model. The example I linked about for how to customize a backend is the one on the main python social auth site.

dctrud:
> I am using a custom user model, and django social auth with pipelines, you can read more about here --> https://python-social-auth.readthedocs.io/en/latest/pipeline.html#extending-the-pipeline.

Yeah, I've read that. django_auth_ldap will use the custom user model and I don't think the pipeline is an issue here. I'll carry on and get something a bit more separated working.

vsoch:
sounds good

dctrud:
@vsoch - I've now pulled out almost everything into `shub.plugins.ldap_auth`. Have played around and come up with a plugin structure where there is a list `PLUGINS_ENABLED` in config.py and things listed there will be added to `INSTALLED_APPS`, have their urls integrated etc. 

Am I on the right track for your thinking here?

NB - the agreement to terms is working here. By making the ldap-login go to the standard social-auth /login endpoint on success we then hook in to the terms check there.

dctrud:
Title has now been changed as this is possibly also laying the foundation for a plugin structure that may support e.g. https://github.com/singularityhub/sregistry/issues/14

vsoch:
Yes I completely love it! If you can give me enough guidance to set it up, I'll do a PR to your PR with some additions (e.g., we can do the same kind of structure for the other login types) and I can give a first go at a doc that outlines "How to Contribute a Plugin"

vsoch:
hey sorry I didn't get to review this yet over the weekend - I'm behind on a few things. I should (hopefully) be able to catch back up this week!

dctrud:
Okay - there are some rudimentary docs and a config example here now. LDAP directories are generally pretty unique though. I tend to just refer to broader django-auth-ldap guide than want to provide lots of configuration samples for different situations.

docs/plugins.md
```diff
@@ -0,0 +1,53 @@
...
+such as container scanning, LDAP authentication, to be added without complicating the core code of
+sregistry.
+
+Plugins distributed with sregistry are found in the `shub/plugins` directory. 
+
+## Included Plugins
+
+The following plugins are included with sregistry, and can be enabled by adding them to the
+`PLUGINS_ENABLED` entry in `shub/settings/config.py`. Plugins may require further configuration in
+your registries' local `shub/settings/secrets/py` file.
```
vsoch: secrets.py here I think?

docs/plugins.md
```diff
@@ -0,0 +1,53 @@
...
+    `shub/settings/secrets.py.example` for an example OpenLDAP configuration.
+  
+Because no two LDAP directories are the same, configuration can be complex and there are no
+standard settings. The plugin uses `django-auth-ldap`, which provides more [detailed documentation
+at Read the Docs here](https://django-auth-ldap.readthedocs.io/en/1.2.x/authentication.html).
+
+To test LDAP authentication you may wish to use a docker container that provides an OpenLDAP
+directory. `mwaeckerlin/openldap` [(GitHub)](https://github.com/mwaeckerlin/openldap) [(Docker
+Hub)](https://hub.docker.com/r/mwaeckerlin/openldap/) is a useful container configured with SSL and
+
```
vsoch: SSL and... *left hanging*!

vsoch:
Things I'm testing / tweaking:

- [x] Dockerfile image builds ok (yes, had to remove one trailing / after build-essential)
- [x] web interface loads without issue
- [ ] regular site functions (without l-dap) seem to be working
- [ ] LDAP setup following guide (and using Docker image?) I'll write this up if/when I figure it out (never used LDAP before!)
- [ ] LDAP log in alongside other options
- [ ] Make other login options part of a shared variable (like plugins)

vsoch:
heyo @dctrud ! Okay this branch is a clone of this PR, and I've just did a big of reorganization for the plugins docs, eg:

```
docs/
├── assets
├── client.md
├── _config.yml
├── credentials.md
├── faq.md
├── install.md
├── introduction.md
├── *plugins*
│   ├── ldap.md
│   └── README.md
├── README.md
├── setup.md
├── singularity-client.md
└── use-cases.md
```

Here is the start of what I think we want to do - have a very brief "here is how to get it working" with the Docker image. I started to try things out, and got a bit lost because I've never used LDAP.

https://github.com/singularityhub/sregistry/blob/dctrud-ldap_auth/docs/plugins/ldap.md

Could you write out the steps you would take (with explanation) for deploying the Docker example? You can just bullet in an issue if you want and I'll add verbosity in the doc. Once we have that working I can test it with the rest of the site, and when I merge that branch it will merge the PR. Woot!

dctrud:
@v - could you work directly on this ldap_auth PR branch instead? It's setup to allow edits from maintainers and is much cleaner to see the history progress if you work directly on it as it'll all come up nicely here. I can merge back your dct-ldap_auth but it might get messy.

vsoch:
yeah! Let me see if I can figure out how to do that, heh.

dctrud:
> yeah! Let me see if I can figure out how to do that, heh.

Meh - I deleted last comment as it's actually not hard to merge - it does come out neatly here. :-)

vsoch:
oh okay I was about to push! I'll hold off.

vsoch:
I've never done that before, I'm used to pulling the branch, making a few trivial tweaks, and then just merging it.

dctrud:
@vsoch - see 3dee864 for a very quick go at some docker ldap testing docs. *Hopefully* this works for you. I will try to go through it also some evening this week, so I can be sure I didn't miss anything. I'm afraid I was writing as I go, and don't have time right now to go back and test end to end.

vsoch:
got it! And merged it into this branch again. Don't worry about testing - I should be able to do a bunch onc with your guidance on the LDAP.

vsoch:
I'm having some trouble with authenticating the Docker image - I posted an issue in case there is a bug in the start.sh script --> https://github.com/mwaeckerlin/openldap/issues/1

vsoch:
I surprisingly did try that one too! Here is what I get:

```
root@docker[d87cacf9a835]:/# ldapwhoami 
SASL/DIGEST-MD5 authentication started
Please enter your authentication name: cn=config,dc=my-example,dc=com    
Please enter your password: 
ldap_sasl_interactive_bind_s: Invalid credentials (49)
	additional info: SASL(-13): user not found: no secret in database
                                                                                                                                             24.10.17 17.51.16
root@docker[d87cacf9a835]:/# ldapwhoami -x
anonymous
```

I am shelling in in the first place because the cat of the example file produced this invalid credential error.

dctrud:
> `Please enter your authentication name: cn=config,dc=my-example,dc=com`

Should be `cn=admin,dc=my-example,dc=com` I believe.

vsoch:
Nope this does not work.

vsoch:
https://asciinema.org/a/i7flmSSBzlm2pcPXwQc86SXzO?speed=3

dctrud:
Okay - this is because it's still trying a SASL/DIGEST-MD5 bind in those cases, and the setup isn't complete for that. There's much setup involved (http://www.openldap.org/doc/admin24/sasl.html)

Use simple authentication only, like in the ldapadd and ldapsearch commands in my docs. Use '-x' and specify the DN to bind with '-D' and that you'll provide a password '-W'

capture

dctrud:
The two small commits above fix one biohpc dn I missed changing to my-company, and add a logging handler to expose LDAP log messages in the log output of the uwsgi app.

If I run sregistry via docker-compose the logs then show that the uwsgi can't access the LDAP server. I guess you are running the sregistry app itself from docker-compose? I forgot I wasn't - I was running from host python, and therefore it could access localhost:389 which is where the LDAP docker directory is exposed. Running the uwsgi under docker-compose would need it to link the openldap container, and the access ldap://openldap I think. Sorry about this. I'm always running sregistry on native python, or via a modified docker-compose so I missed this issue.

Do you want to instruct users to modify their docker-compose.yml to bring LDAP inside that, or some other approach for the example?

Again sorry about this - I almost never use a distributed docker compose as-is, since I'm integrating with an existing multi-app nginx-proxy/letsencrypt setup, and other stuff.

vsoch:
Hey no worries we are very close! I was actually running it outside of docker-compose (as it's own separate container) since it was just an example to get started. That's the most basic example, so let's stick with that, and we can suggest to add to docker-compose if they need. But I would bet most already have their ldap setup and just want to connect to it. For logging - any method that shows the user how to debug and look at logs would be good (the user being, me, haha).

dctrud:
django-auth-ldap writes no logs anywhere by default. The stuff in 5af8c32 makes sure it will go into the main stream of messages from the uwsgi process. In that change I set it to DEBUG level, but arguably INFO could also be OK.

vsoch:
boom boom! It worked beautifully! I added a few more notes about logging, and also noted that the user needs to inspect the running openldap container to get the url (I got a message that the server wasn't found with 127.0.0.1. 

I'm loving the new plugins framework - and the LDAP login! This is great work @dctrud !

PR #52, opened 2017-10-31
updating version of singularity python

vsoch:
@victorsndvg this PR is coming after updating the requirements.txt to use a (newly merged) singularity python which should have the correct install and not the date issue!
If you have some bandwidth, could you try testing out with your build? A clean install should grab the right one now. Definitely no rush - looking forward to hear if it works, and (I still) would like to add that nice install script that you are working on (when you think it's ready) :)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -33,7 +33,7 @@ django-chosen
 opbeat
 django-hstore==1.3.5
 django-datatables-view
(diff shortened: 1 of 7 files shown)


vsoch:
This is confirmed to work by @victorsndvg and also has addition to fix a bug with container deletion. The imageFile needed to be deleted pre_save instead of post_save.

PR #69, opened 2017-12-16
Update/mpl2.0

vsoch:
@dctrud and @akkornel here is a preview of what the GPL 2.0 would look like, and we might want to write out something in humanspeak for the CONTRIBUTING.md.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,21 +1,5 @@
-# Contributor's Agreement
-
-You are under no obligation whatsoever to provide any bug fixes, patches,
-or upgrades to the features, functionality or performance of the source
-code ("Enhancements") to anyone; however, if you choose to make your
-Enhancements available either publicly, or directly to the project,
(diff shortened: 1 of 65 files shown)


dctrud:
Something along the lines of the tldrlegal blurb?

https://tldrlegal.com/license/mozilla-public-license-2.0-(mpl-2)

>You must make the source code for any of your changes available under MPL, but you can combine the MPL software with proprietary code, as long as you keep the MPL code in separate files. Version 2.0 is, by default, compatible with LGPL and GPL version 2 or greater. You can distribute binaries under a proprietary license, as long as you make the source available under MPL.

vsoch:
Oh I like that! I think we should be having some conversation about this in the new year, will keep you posted! :fist_oncoming: <-- Happy Boxing Day!

vsoch:
Arching this PR as it is no longer needed, and we can re-open / re-visit if it is!

PR #77, opened 2018-01-06
Allowing for private (authenticated pull) of images

vsoch:
This PR will do the following:
 - updates the sregistry to use the sregistry client from this this PR: https://github.com/singularityhub/sregistry-cli/pull/32 Note that although the PR isn't merged, the client version (0.0.5) has been pushed to pypi. @victorsndvg when you test, make sure to install this verison 0.0.5 from pypi.
 - **important** the structure of the credential provided has been changed (to allow for the other clients). Instead of a flat file:

```
{'token': '1234' ..}
```
it's now nested under registry

```
{ 'registry'' { token': '1234' .. } }
```
so @victorsndvg - you will need to go to the web interface to get the updated token data structure, upload it at $HOME/.sregistry (and this is the format that this PR for sregistry and the client version 0.0.5 are expecting).

@victorsndvg would you like to test this branch and then report back? Specifically, we should do the following:

1. pull this branch and use it to update your local registry
2. go to the token page in the registry and copy paste the newly formatted token to your `$HOME/.sregistry` file
3. Install the version of sregistry from pip (should be 0.0.5) both locally for your client, and for your containers (or for any singularity images you use with the client).
4. test the pull, report back

When we have confirmed all looks good, I'll merge both PRs! Note that although sregistry 0.0.5 is on pypi it hasn't been merged either.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and changes prior to that are (unfortunately) done retrospectively. Critical ite
 ## [v1.x](https://github.com/singularityhub/singularity-python/tree/development) (development)
 
 **changed defaults**
+ - changed internal functions to use sregistry client (instead of singularity)
+ - authentication check added to sregistry pull, so private images can be pulled given correct credentials
  - from the *sregistry client* provided by Singularity Python, to support use of squashfs images and singularity 2.4, the default upload is not compressed, assuming squashfs, and the default download is not decompressed. To still compress an image add the `--compress` flag on push, and the `--decompress` flag on pull.
 
 **bugs fixed**
--- a/requirements.txt
+++ b/requirements.txt
@@ -33,7 +33,7 @@ django-chosen
 opbeat
 django-hstore==1.3.5
 django-datatables-view
--e git+https://github.com/singularityware/singularity-python.git@1b38684fcda7931885269488e16305f989147b4f#egg=singularity
+sregistry
 django-gravatar2
 pygments
 google-api-python-client
--- a/requirements.txt
+++ b/requirements.txt
@@ -33,7 +33,7 @@ django-chosen
 opbeat
 django-hstore==1.3.5
 django-datatables-view
-sregistry
+sregistry==0.0.5
 django-gravatar2
 pygments
 google-api-python-client
--- a/shub/apps/api/actions/delete.py
+++ b/shub/apps/api/actions/delete.py
@@ -20,7 +20,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
 '''
 
 from shub.logger import bot
-from singularity.registry.auth import generate_timestamp
+from sregistry.auth import generate_timestamp
 from shub.apps.api.utils import validate_request
 
 
--- a/shub/apps/api/actions/push.py
+++ b/shub/apps/api/actions/push.py
@@ -27,7 +27,7 @@ from rest_framework.viewsets import ModelViewSet
 from shub.apps.api.models import ImageFile
 from rest_framework import serializers
 from shub.apps.api.utils import validate_request
-from singularity.registry.auth import generate_timestamp
+from sregistry.auth import generate_timestamp
 
 class ContainerPushSerializer(serializers.HyperlinkedModelSerializer):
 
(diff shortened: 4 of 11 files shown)


victorsndvg:
Hi @vsoch,

it works!

I successfully download an image from a private collection owned by myself. I figure out I will not be able to download an image from a collection owned by others. I'm right?

I also did a weird test ... I removed the credentials file (`.sregistry` located at $HOME) and then tried to pull this private image. I know this is not the right way to do it, but I think `sregistry-cli` should manage this. The error message i get was:

```
$ singularity run --app registry sregistry-cli_0.0.5 pull mso4sc/sregistry-cli:latest
[client|registry] [database|sqlite:////home/vsande/.singularity/sregistry.db]
Traceback (most recent call last):
  File "/opt/conda/bin/sregistry", line 11, in <module>
    load_entry_point('sregistry', 'console_scripts', 'sregistry')()
  File "/opt/sregistry-cli/sregistry/client/__init__.py", line 319, in main
    subparser=subparsers[args.command])
  File "/opt/sregistry-cli/sregistry/client/pull.py", line 39, in main
    save=do_save)
  File "/opt/sregistry-cli/sregistry/main/registry/pull.py", line 62, in pull
    manifest = self._get(url)
  File "/opt/sregistry-cli/sregistry/main/base.py", line 214, in _get
    return_json=return_json)
  File "/opt/sregistry-cli/sregistry/main/base.py", line 322, in _call
    stream=stream)
  File "/opt/conda/lib/python3.6/site-packages/requests/api.py", line 72, in get
    return request('get', url, params=params, **kwargs)
  File "/opt/conda/lib/python3.6/site-packages/requests/api.py", line 58, in request
    return session.request(method=method, url=url, **kwargs)
  File "/opt/conda/lib/python3.6/site-packages/requests/sessions.py", line 508, in request
    resp = self.send(prep, **send_kwargs)
  File "/opt/conda/lib/python3.6/site-packages/requests/sessions.py", line 612, in send
    adapter = self.get_adapter(url=request.url)
  File "/opt/conda/lib/python3.6/site-packages/requests/sessions.py", line 703, in get_adapter
    raise InvalidSchema("No connection adapters were found for '%s'" % url)
requests.exceptions.InvalidSchema: No connection adapters were found for 'None/container/mso4sc/sregistry-cli:latest'
```
I think a more friendly error message should be shown.
What do you think?

vsoch:
Woohoo! I totally agree about providing a more user friendly message, and would you want only the owner to be able to download the image? I think we likely need some other level of groups / permissions for this particular case of pulling private. I think these are the most likely use cases:

- private and limit pull to collection owner
- private but allow anyone to pull that knows the image uri
- private and allow pull to some select group of registry users decided by owner

I was thinking the second is most likely wanted - images aren't known but discovered easily if you know they exist. This would add one setting to limit private container pull to just collection owners (or not). What do you think?

FYI - currently still traveling! When I am able I will update the PR with some new fixes for testing and ping you after.

victorsndvg:
Hi @vsoch ,

Let me know any improvement, I'm your beta tester! :)

I disagree. I think the second alternative is not secure enough. As an example, think in a private source code repository, it can contains secret data, private/licensed software, etc. You, as the owner, want to control who has access to it, grant permissions, etc.  Applying the second alternative means that something can access your top-secret things by mistake (or brute force attacks) ... :thinking: 

About the first alternative, I think is the most restrictive. By default an owner cannot share the collection. If a collection owner want to share his/her collection he/she has to share the secret token. Probably is not the best option ... but he/she is the only with power to do that.

I prefer the third option. The collection owner has the power to share the images and, in addition, there is a central point where to "formally" store who can access to a private collection. 

In my opinion, mixing 1st and 3rd option is the best solution. This means:
- A collection can be owned by several users (3rd option). 
- Collection owners can add/remove other owners (3rd option). 
- Only owners can download a container from a private collection (like in 1st option)

What do you think?

Have a nice travel!

vsoch:
I agree, I think that we need to have more advanced permissions / ownership for collections. I don't think I have time to implement this soon, but I'll do it after I recover!

victorsndvg:
Ok, note that I'm writing as soon as I see your comments, but I'm never waiting for a quick response.

Please, take the time you need to recover!!

Best wishes!

vsoch:
hey @victorsndvg if this basic function to allow for private pull is ok, let's get the changes merged to not get too mixed into the others. We can open a separate issue and then PR for the remainging user roles, and get @dctrud feedback as well.

victorsndvg:
Sorry, I forgot this last message...

For me it's ok. Great job!

Issue #73, opened 2017-12-27
Default roles for new users


victorsndvg:
Hi @vsoch ,

SRegistry is taking advantage of `python-social-auth` and comes with several  SSO preconfigured providers like Twitter, Google, etc.

In the particular case of `Fiware` (and probably others) you can locally install the identity manager. This means that the SRegistry admin/deployer can control/manage the user accounts under a particular authentication strategy. This is something similar to LDAP, but providing SSO.

In the project we are using SRegistry, we need to manage users and we also need SSO capabilities. This why we are going to use only `Fiware` and `LDAP` for authentication. We are going to use these authentication strategies with different purposes, `Fiware` to manage the accounts of a new portal containing several services and requiring SSO, and `LDAP` for the company.

We want to assign different  (or the same) default privileges (roles) depending on the authentication method used. I would like to assign the admin role as default for `Fiware` users and superuser (or admin) role for `LDAP` users.

In my mind I have two possible solutions,
1.- Assign a default role for all users. E.g. all users are admin as default
2.- Assign a default role per OAuth strategy. E.g. all `Fiware` users are admin and all `LDAP` users are superuser as default.

What do you think? It has sense from your point of view?

vsoch:
hey @victorsndvg ! I think sregistry should have basic roles to determine admin vs not (the manager and admin as exists now) and then have the option for the user to create additional roles for users (whatever they may be, nothing hard coded) using standard django permissions. https://timonweb.com/posts/how-to-get-a-list-of-all-user-permissions-available-in-django-based-project/ So we should chat about a way to support that while still being flexible to 

 1. not having any extra types at all, and 
 2. having one or more custom types that are added in a config somewhere, and then editable both from manage.py and possibly interface
 
It might be best to look into packages that manage permissions, or to take the most basic implementation and turn it into a plugin. Let me know your ideas for that!

victorsndvg:
Hi @vsoch ,

note that I don't need extra roles. What I desire is a way to assign a default role for first-time-loged-in users instead of using the manage.py CLI each time a new user wants to get the right access to manage collections

vsoch:
[...]

PR #83, opened 2018-01-25
removing admin role in favor of is_staff

vsoch:
This will remove the `admin` boolean stored with a user in favor of the standard `is_staff` which is a much better option, since it's a django default. See here https://github.com/singularityhub/sregistry/issues/73#issuecomment-360592999 for discussion, and @dctrud and @victorsndvg looking for your feedback on this!

### DIFF
--- a/shub/apps/api/utils.py
+++ b/shub/apps/api/utils.py
@@ -95,7 +95,7 @@ def validate_request(auth,
         return False
 
     if superuser is True:
-        if user.admin is False:
+        if user.is_staff is False:
             bot.debug('User %s is not a superuser, request invalid.' %user.username)
             return False
 
--- a/shub/apps/base/templates/base/navigation.html
+++ b/shub/apps/base/templates/base/navigation.html
@@ -48,7 +48,7 @@
                 <li role="separator" class="divider"></li>
                 <!--<li class="dropdown-header">Settings</li>-->
                 <li><a href="/api">API</a></li>
-                {% if request.user.admin %}
+                {% if request.user.is_staff %}
                 <li><a href="{% url 'token' %}">Token</a></li>
                 {% endif %}
                 <li><a href="{% url 'logout' %}">Log Out</a></li>
--- a/shub/apps/main/models.py
+++ b/shub/apps/main/models.py
@@ -75,7 +75,7 @@ def has_edit_permission(instance,request):
         return False
 
     # Global Admins
-    if request.user.admin is True:
+    if request.user.is_staff is True:
         return True
 
     if request.user.is_superuser is True:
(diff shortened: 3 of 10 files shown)


dctrud:
HI @vsoch - this looks good, except I think it might be the time now to start committing the migrations in the repository, rather than always making them on the fly in the sregitry startup script. With a manual customized migration it would be possible to take the existing values for each user.admin value and put into user.is_staff

Using on-the-fly make migrations in production is a bit scary, as you can't inspect the repo to check what's going to happen to existing values when there are schema changes like this one.

dctrud:
Understood. Actually for LDAP users like us, who are configuring a group to set `is_staff`, then it won't be an issue without a custom migration - since the LDAP integration will set is_staff appropriately on login. Is just going to affect people not using that - who have to reset people to be staff.

dctrud:
This PR is working for me. With the following LDAP config set:

```
# Map LDAP group membership into Django admin flags
AUTH_LDAP_USER_FLAGS_BY_GROUP = {
    "is_superuser": "cn=sregistry_admin,ou=Group,ou=users,dc=example,dc=com",
    "is_staff": "cn=sregistry_admin,ou=Group,ou=users,dc=example,dc=com",
}
```

I have token access, so I'm picked up as an admin, through is_staff by virtue of my LDAP group membership.

vsoch:
okay awesome! Question - what do you think about the is_active status? Should we keep it in the LDAP example (as we have now) even though it's not used anywhere? Or just nix it? I was trying to think of if there are good places for having active vs. non active users, but didn't think of anything compelling.

dctrud:
It is often safe to remove the `is_active`. If you want to limit login to sregistry to a subset of users it's often possible to do this via LDAP query in`AUTH_LDAP_USER_SEARCH`

The `is_active` is likely of most use if you don't archive inactive user accounts to a sub-tree outside of your active users tree in LDAP. Then you might want to limit logins to active users, who are part of an active users group.

On our LDAP directory we remove inactive accounts to a seperate OU (subtree) that is outside of the one for active accounts. AUTH_LDAP_USER_SEARCH is set to look only at the active user OU (subtree) so inactive accounts cease to be visible to the sregistry app, and can't login at all.

vsoch:
okay, I think I'll conceal it from the setup for now.

vsoch:
okay all set. We can add the variable for https verify=False here if you like too, saves the additional PR and messing something up with Github :)

vsoch:
Do you think there should be a little blurb in the LDAP docs to explain the mapping of roles?

dctrud:
The https verify thing is for the client though, not the sregistry server?

Mapping of roles blurb is difficult. Like much stuff with LDAP, how you use it depends on your directory structure. I think the comments in the settings example are probably good.

vsoch:
Yes, don't mind me, dinosaur brain fart. Merging! Thanks for your speedy help to fix this up :)

PR #85, opened 2018-01-26
Fixing Gitlab links and documentation

vsoch:
This PR will do the following:
 - update the url to reference "gitlab" and not the oauth 2 (which seems to be correct)
 - add the backend to the list (it was missing)


vsoch:
Backend enabled:

image

Redirect to login:

image

Back to terms!

image


We are good :)

image

Issue #82, opened 2018-01-18
Connection timeout pulling big containers


victorsndvg:
Hi @vsoch,

Sorry again for bringing this nginx errors ... but I'm not very experienced with this web services and I think this could be easy for you.  :P

I'm trying to pull a 2.3Gb image and I always get the following error:
```
nginx_1   | 2018/01/18 16:46:59 [error] 6#6: *3 upstream timed out (110: Connection timed out) while reading response header from upstream, client: XXXXXX, server: localhost, request: "GET /containers/8/download/3c548175-fd5c-4033-b667-be836f0c7f4b HTTP/1.1", upstream: "uwsgi://172.17.0.4:3031", host: "XXXXXXX2
```
The connections are  always closed after 60s.

This is the current nginx.conf file

```
server {
  listen                *:80;
  server_name           localhost;

  client_max_body_size 8000M;
  client_body_buffer_size 2000M;
  client_body_timeout 900;
  send_timeout 900;

  add_header X-Clacks-Overhead "GNU Terry Pratchett";
  add_header X-Clacks-Overhead "GNU Terry Pratchet";
  add_header Access-Control-Allow-Origin *;
  add_header 'Access-Control-Allow-Credentials' 'true';
  add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS';
  add_header 'Access-Control-Allow-Headers' 'Authorization,DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type';

  location /images {
    alias /var/www/images;
  }

  location / {
    include /etc/nginx/uwsgi_params.par;
    uwsgi_pass uwsgi:3031;
  }

  location /static {
    alias /var/www/static;
  }
}

server {

        listen 443;
        server_name localhost;

        root html;
        client_max_body_size 8000M;
        client_body_buffer_size 2000M;
        client_body_timeout 900;
        send_timeout 900;

        add_header X-Clacks-Overhead "GNU Terry Pratchett";
        add_header X-Clacks-Overhead "GNU Terry Pratchet";
        add_header Access-Control-Allow-Origin *;
        add_header 'Access-Control-Allow-Credentials' 'true';
        add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS';
[...]

PR #90, opened 2018-02-02
Address #82 large container pulls with download streaming

dctrud:
This PR is to address #82 - there are 3 changes:

- 682bba3 sets up the Django logger to direct WARNING and above to console by default, so Django errors can then be seen with `docker-compose logs uwsgi`. This is critical to effectively debug download issues that are occuring in Django code. The logs coming out of UWSGI alone are not enough.

- 3690bac uses the Django `FileResponse` class to serve a container download without requiring it to be read entirely into RAM. Note that underneath this is going to use `wsgi.filewrapper` as we are running via UWSGI. `FileResponse` is a convenience (vs using the filewrapper and StreamingHttpResponse directly) that was introduced in newish Django versions. We can't use a `with ... as f` context yet, as context manager support for `File.open` was added in Django 2.0 and we are using 1.11

- bf1e505 disables harakiri for UWSGI. Since we are serving large files via the app/UWSGI we don't know how long is reasonable for a request - there may be a slow connection, a multi GB download could take hours. Auto-kill of unusually long running UWSGI reqests would be appropriate if we further offload serving downloads using xsendfile so that the download itself is handled by nginx, and not our UWSGI  app.

### DIFF
--- a/shub/apps/main/views/download.py
+++ b/shub/apps/main/views/download.py
@@ -99,18 +99,11 @@ def download_share(request,cid,secret):
     if secret != share.secret:
         raise Response(status.HTTP_401_UNAUTHORIZED)
 
-    # If we make it here, link is good
-    filename = container.get_download_name()
-
-    f = open(container.image.get_abspath(), 'rb')
-    response = FileResponse(f, content_type='application/img')
-    response['Content-Disposition'] = 'attachment; filename="%s"' %filename
-    
-    return response
+    return _download_container(container)
 
 
 
-def download_container(request,cid,secret):
+def download_container(request, cid, secret):
     '''download a container
     '''
     container = get_container(cid)
@@ -119,10 +112,28 @@ def download_container(request,cid,secret):
     if container.secret != secret:
         return Http404
 
-    filename = container.get_download_name()
+    return _download_container(container)
 
-    f = open(container.image.get_abspath(), 'rb')
+
+def _download_container(container):
+    '''
+       download_container is the shared function between downloading a share
+       or a direct container download. For each, we create a FileResponse
+       with content type application/img, and stream it to the container's
+       download name. A FileResponse is returned.
+
+       Parameters
+       ==========
+       container: the container to download
+
+    '''
+
+    filename = container.get_download_name()
+    filepath = container.image.get_abspath()
+
+    f = open(filepath, 'rb')
     response = FileResponse(f, content_type='application/img')
     response['Content-Disposition'] = 'attachment; filename="%s"' %filename
+    response['Content-Length'] = os.path.getsize(filepath)
 
     return response
--- a/shub/apps/main/views/download.py
+++ b/shub/apps/main/views/download.py
@@ -35,7 +35,8 @@ from django.shortcuts import (
 
 from django.http import (
     JsonResponse,
-    HttpResponse
+    HttpResponse,
+    FileResponse
 )
 
 from shub.apps.main.utils import (
@@ -100,9 +101,11 @@ def download_share(request,cid,secret):
 
     # If we make it here, link is good
     filename = container.get_download_name()
-    response = HttpResponse(container.image.datafile.file,
-                            content_type='application/img')
+
+    f = open(container.image.get_abspath(), 'rb')
+    response = FileResponse(f, content_type='application/img')
     response['Content-Disposition'] = 'attachment; filename="%s"' %filename
+    
     return response
 
 
@@ -117,7 +120,9 @@ def download_container(request,cid,secret):
         return Http404
 
     filename = container.get_download_name()
-    response = HttpResponse(container.image.datafile.file,
-                            content_type='application/img')
+
+    f = open(container.image.get_abspath(), 'rb')
+    response = FileResponse(f, content_type='application/img')
     response['Content-Disposition'] = 'attachment; filename="%s"' %filename
+
     return response
--- a/shub/settings/logging.py
+++ b/shub/settings/logging.py
@@ -1,0 +1,17 @@
+# Default Django logging is WARNINGS+ to console
+# so visible via docker-compose logs uwsgi
+LOGGING = {
+    'version': 1,
+    'disable_existing_loggers': False,
+    'handlers': {
+        'console': {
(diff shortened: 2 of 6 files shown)


vsoch:
That was so speedy! Testing now.

shub/settings/main.py
```diff
@@ -128,7 +128,29 @@
 GRAVATAR_DEFAULT_IMAGE = "retro"
 # An image url or one of the following: 'mm', 'identicon', 'monsterid', 'wavatar', 'retro'. Defaults to 'mm'
 
+
+# Default Django logging is WARNINGS+ to console
+# so visible via docker-compose logs uwsgi
+LOGGING = {
+    'version': 1,
```
vsoch: tiny detail - let's be overly organized and put this in another settings file called `logging.py` so the developer finds it immediately!

vsoch: Actually if I merge from a push (after @victorsndvg and I test) I can do this easily.

dctrud: Ahh ... I already ... see below :-)

shub/settings/logging.py
```diff
@@ -0,0 +1,17 @@
+# Default Django logging is WARNINGS+ to console
+# so visible via docker-compose logs uwsgi
```
vsoch: I think this guy needs an `import os`

dctrud:
@vsoch - you can't do with ```open... as f:``` as the actual file handling via wsgi.filewrapper doesn't occur in that context - if you try that it will error out in the wsgi code. The wsgi FileWrapper will call close()

See: https://docs.python.org/2/library/wsgiref.html

AFAIK once you have Django 2.0 you can maybe do it a little more elegantly given the context manager support on the File class.

shub/settings/logging.py
vsoch: and the logging object

vsoch:
Ah ok. That's very cool!

shub/settings/logging.py
dctrud: logging object is not needed - we aren't calling it. all the 'logging.' things are strings.

dctrud:
Ultimately using xsendfile is really the best way, so your uwsgi processes aren't tied up dealing with pushing bytes from files down the wire. But... this is quicker to implement. xsendfile setup needs furthere nginx configuration too and a little care to ensure you aren't going to expose anything via senfile to a client who shouldn't get it.

vsoch:
Ok here is a log of testing!

1. I cloned the repository, built and did `docker-compose up -d`
2. An import of `os` needed to be added to logging
3. `docker-compose restart`
4. Log in with twitter
5. Then I shelled into the main and added myself as an admin and superuser

```
docker exec -it 5ce7010a2f99 bash
$ python manage.py add_superuser --username vsoch
DEBUG Username: vsoch
DEBUG vsoch is now a superuser.
$ python manage.py add_admin --username vsoch
DEBUG Username: vsoch
DEBUG vsoch can now manage and build.
```
Then I downloaded my credentials at `http://127.0.0.1/token`, updated them at `$HOME/.sregistry` and used singularity global client to push an image:

```
export SREGISTRY_CLIENT=registry
registry sregistry push --name peanut-butter/jelly:time vsoch-hello-world-master-latest.simg 
[client|registry] [database|sqlite:////home/vanessa/.singularity/sregistry.db]
[================================] 62/62 MB - 00:00:00
[Return status 201 Created]
```
Search...
```
sregistry search
 sregistry pull peanut-butter/jelly:time
[client|registry] [database|sqlite:////home/vanessa/.singularity/sregistry.db]
[container][update] peanut-butter/jelly:time
Success! /home/vanessa/.singularity/shub/peanut-butter-jelly:time.simg
$ sregistry get peanut-butter/jelly:time
/home/vanessa/.singularity/shub/peanut-butter-jelly:time.simg
$ singularity run $(sregistry get peanut-butter/jelly:time)
```
Ah yes, so what is missing is the download bar, and this is because the Content-Size is not there. I just added this, and pushed, and I think the PR updated!! I hope I didn't mess anything up. I'm going to try a larger image now.

shub/settings/logging.py
vsoch: yeah I didn't see in the interface that logging.StreamHandler is a string, hehe :) This is why I need my editor colors!

shub/apps/main/views/download.py
```diff
@@ -116,8 +112,28 @@ def download_container(request,cid,secret):
     if container.secret != secret:
         return Http404
 
+    return _download_container(container)
+
+
+def _download_container(container):
```
vsoch: @dctrud see here, since we are doing the same thing in two functions, I  thought this was a good use for a shared ("private" to download.py) function. I also added the content length so that the progress bar comes back, eg here it is with it:

```
sregistry pull peanut-butter/jelly:time
[client|registry] [database|sqlite:////home/vanessa/.singularity/sregistry.db]
Progress |===================================| 100.0% 
[container][update] peanut-butter/jelly:time
Success! /home/vanessa/.singularity/shub/peanut-butter-jelly:time.simg
```

and without is just empty space

```
sregistry pull peanut-butter/jelly:time
[client|registry] [database|sqlite:////home/vanessa/.singularity/sregistry.db]

[container][update] peanut-butter/jelly:time
Success! /home/vanessa/.singularity/shub/peanut-butter-jelly:time.simg
```

vsoch:
okay, I approve! @victorsndvg when you can test with your very large images and sign off, we can merge.

shub/apps/main/views/download.py
dctrud: Yup - LGTM.

vsoch:
@victorsndvg we have two approves and I want to get started on the permissions stuff (with this as a base) so I'm going to merge to master as it's an improvement on the current, and if you have any remaining issues let's open as fresh issues. Merging!

victorsndvg:
Great @dctrud ! 

It works for me,  and there is no increase of RAM memory as you said. Perfect! 👍 

I'm also trying from a veeeeery slow network connection. If I find any new issue I will report.

Thnks agains!

dctrud:
Glad it works, and thanks to @vsoch for fixing up the content-length etc.

PR #96, opened 2018-02-14
[fix collection view]

hansendx:
The button dropdowns in the collection view did not
conform to bootstrap and therefore did not work.
Moved the button end tag.
Moved the button related attributes of the "a" tags surrounding
the button text to the button tags themselves and removed those a tags.
Encapsulated the dropdown buttons in a btn-group div so the dropdown will be placed correctly.


vsoch:
This looks good, thank you!

Issue #106, opened 2018-03-12
why sregistry[-cli] 0.0.5  is needed to build and install?


Trophime:
Hi,
I've noticed in `requirements.txt` needed for installing `sregistry` that you are
using `sregistry[-cli] 0.0.5` from `PiPy`.

I'm wondering why you don't use a more recent version of  `sregistry[-cli]` ?
and why `sregistry-cli` is actually needed?

Trying, very naively to update cli to a recent version, I end up with an error:
```
uwsgi_1   | Traceback (most recent call last):
uwsgi_1   |   File "shub/wsgi.py", line 7, in <module>
uwsgi_1   |     application = get_wsgi_application()
uwsgi_1   |   File "/usr/local/lib/python3.5/site-packages/django/core/wsgi.py", line 13, in get_wsgi_application
uwsgi_1   |     django.setup(set_prefix=False)
uwsgi_1   |   File "/usr/local/lib/python3.5/site-packages/django/__init__.py", line 27, in setup
uwsgi_1   |     apps.populate(settings.INSTALLED_APPS)
uwsgi_1   |   File "/usr/local/lib/python3.5/site-packages/django/apps/registry.py", line 108, in populate
uwsgi_1   |     app_config.import_models()
uwsgi_1   |   File "/usr/local/lib/python3.5/site-packages/django/apps/config.py", line 202, in import_models
uwsgi_1   |     self.models_module = import_module(models_module_name)
uwsgi_1   |   File "/usr/local/lib/python3.5/importlib/__init__.py", line 126, in import_module
uwsgi_1   |     return _bootstrap._gcd_import(name[level:], package, level)
uwsgi_1   |   File "<frozen importlib._bootstrap>", line 986, in _gcd_import
uwsgi_1   |   File "<frozen importlib._bootstrap>", line 969, in _find_and_load
uwsgi_1   |   File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
uwsgi_1   |   File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
uwsgi_1   |   File "<frozen importlib._bootstrap_external>", line 662, in exec_module
uwsgi_1   |   File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
uwsgi_1   |   File "./shub/apps/api/models.py", line 24, in <module>
uwsgi_1   |     from shub.apps.api.actions import create_container
[...]

PR #107, opened 2018-03-12
updating sregistry to version 0.0.74

vsoch:
This PR will update the sregistry version, and specify installation of the registry module to ensure requests_toolbelt and the timestamp generation functions are installed. Notably, we in the api and url functions we move the timestamp generation function into the registry module as they are no longer in a global auth.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,9 @@ represented by the pull requests that fixed them. Critical items to know are:
  - changed behaviour (recipe sections work differently)
 
 
-## [v1.x](https://github.com/singularityhub/singularity-python/tree/master) (master)
+## [master](https://github.com/singularityhub/singularity-python/tree/master) (master)
 
+ - updating sregistry-cli to version 0.0.74
  - superusers and admins (global) can now create collections via a button in the interface
  - demos and customizated supplementary content removed - not used
  - user customization possible by superusers in the admin panel
--- a/docs/pages/setup.md
+++ b/docs/pages/setup.md
@@ -183,6 +183,10 @@ Tacosaurus Computing Center is a Singularity Registry to provide institution lev
 
 ```
 
+At this point, you can send this file to `@vsoch` and she will be happy to add your 
+registry to the... registry! If you want to customize your robot image, or submit
+the file yourself via a pull request, continue reading!
+
 ### 3. Choose your image
 For the image and thumbnail, we have a [database of robots](https://vsoch.github.io/robots) that we have randomly selected a robot for you. If you don't like your robot, feel free to browse and choose a different one. Importantly, you will need to add the robot to the github repo:
 
--- a/docs/pages/setup.md
+++ b/docs/pages/setup.md
@@ -144,7 +144,9 @@ cd containers
 ```
 
 ### 2. Generate your Metadata
-Then use the manager to generate a markdown file for your registry:
+Then use the manager to generate a markdown file for your registry. In this example, we have 
+interactively shelled inside the `uwsgi` image for our sregistry like `docker run -it <container> bash` 
+and are sitting in the `/code` folder.
 
 ```
 # Inside the image
@@ -153,7 +155,7 @@ $ python manage.py register
 Registry template written to taco-registry.md!
 
 Your robot is at https://vsoch.github.io/robots/assets/img/robots/robot5413.png
-1. Fork and clone https://www.github.com/singularityhub/sregistry
+1. Fork and clone https://www.github.com/singularityhub/containers
 2. Add taco-registry.md to the registries folder
 3. Download your robot (or add custom institution image) under assets/img/[custom/robots]
 4. Submit a PR to validate your registry.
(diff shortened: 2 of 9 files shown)


vsoch:
This will close #106 - @Trophime would you like to take a look? Are you ok with updating the version and still being specific, so we can still be sure about future changes? I've tested locally as well. When you approve we can merge into master.

Trophime:
Before testing the PR, I've tried to move from http to https as:
`singularity pull shub://localhost/...` seems to requires https.

I'm having some troubles when trying to get https working (with twitter oauth so far)
Do we need a valid certificate  (I mean not self signed) for https to work?

vsoch:
If you need regular http to work, then you should check the `DOMAIN_NAME` in settings (see here https://github.com/singularityhub/sregistry/blob/master/shub/settings/config.py#L42), along with making sure you are using the correct docker-compose and nginx.conf (both in this folder https://github.com/singularityhub/sregistry/tree/master/https) that expose port 443. For twitter, (and most oauth) the callback URL is also really important! So if there is a difference it can reject. If the error is on the redirect, then sometimes you can just refresh and it will be logged in anyway. :)

Trophime:
As I have a trouble to get a certificate, I revert to http.
So locally everything works smoothly ;)
However on a remote machine I've tried to pull an image using `sregistry pull salome/8.4.0:buster`:
```
[client|registry] [database|sqlite:////home/LNCMI-G/trophime/.singularity/sregistry.db]
Progress |===========================|-------|  79.0% 
ERROR Return Code 255: ERROR  : Failed to mount squashfs image in (read only): Invalid argument
ABORT  : Retval = 255

WARNING Inspect command not supported, metadata not included.
[container][new] salome/8.4.0:buster
Success! /home/LNCMI-G/trophime/singularity_images/salome-8.4.0:buster.simg
```
and trying to run the container fails:
```
singularity shell /home/LNCMI-G/trophime/singularity_images/salome
ERROR  : No valid /bin/sh in container
ABORT  : Retval = 255
```

vsoch:
If you are testing on a remote machine, can you share with me the url so I can reproduce? and what sregistry version you are using would be helpful too.

Trophime:
my sregistry is installed on calcul18.lcmi.local and I use singularity 2.4.4-1 and sregistry-cli 0.0.74 on the remote machine. For giving you access to the sregistry I would need your IP to open appropriate port and you have to connect with twitter... and send you back the IP of calcul18

Trophime:
if there a limit when downloading with http? 
because it seems to work for "small" images (eg `1008M images/hifimagnet/latestbuster.simg`)
but not for `1.5G images/ensight/10.2.2mso4sc.simg`

vsoch:
ah that might be it! The limit is set by the nginx header, and it looks like we might have missed updating one (here) https://github.com/singularityhub/sregistry/blob/master/https/nginx.conf.https#L36 for the https. It would be good to try changing that, restarting, and see if it fixes. What seems curious is that it tries to do something before reaching 100% and then has an invalid argument issued. So it may not be the header.

Trophime:
but we have for http in `nginx.conf`:
```
  client_max_body_size 8000M;
  client_body_buffer_size 8000M;
```
these values exceed limits set for https no?

vsoch:
yes but if you are using https it will use the https server block I think.

Trophime:
I've setup everything for using http not https at least that what I think
How can I be sure?

Trophime:
Actually, there is no https block defined.
and https port is not open!

vsoch:
Haha, that might be an issue :) See if you can fix that up and try again!

Trophime:
ok I'll give a trial..
I still have a question: can I use a self signed certificate?

vsoch:
If the client allows and the server is configured correctly, I don't see why not. I'm by no means an expert on this though. Generally self signed is good for testing, but I'd advocate for setting up a proper one with lets encrypt. I have done this a few times and my method isn't likely the best, but it works so I'd be glad to help you if needed. I think I documented most of it in the setup for sregistry, so check that first!

Trophime:
I've already did but it seems that I reach some limit of "Lets encrypt" because of my actual domain name "cnrs.fr" ... too bad

vsoch:
I don't think the domain name is issue, I think you need something open (that isn't reachable) to send the tests. This issue seems similar https://community.letsencrypt.org/t/solved-cant-renew-ssl-certificate/55454/16 Domain names have huge variance in terms of extensions, so I don't see how they could limit them. Is there documentation you know of, or more details about your error?

Trophime:
Changing this in `nginx.conf` fix my problem with large file:
```
  location / {
    include /etc/nginx/uwsgi_params.par;
    uwsgi_pass uwsgi:3031;
    uwsgi_max_temp_file_size 10024m;
  }
```

vsoch:
oh this is great news! I didn't even know about that field. Do you want to PR to change it in nginx.conf for the registry (for both http and https?)

vsoch:
@Trophime we got a little distracted with the https :) I've added the tweaks to the nginx.conf to set the max temporary file size - are you able to give the A-OK now on this pull request for merge? It's holding up the other ones!

Trophime:
If we except the pb with https, everything is fine.
You can merge the pull request.
Thanks

vsoch:
Awesome thanks!

PR #117, opened 2018-05-02
Correct link to settings/config.py

jbd:
I don't know if there is a "clean" way to build links between github.io and a github.com repository, so I used the full URI.


jbd:
In fact, most of the relative links in the .md generate a 404. Mayve you'd like to handle it in a certain way ?

vsoch:
You should be reading it in the online interface and not in the markdown files. Eg start at singularity hub.github.io/sregistry. A lot of these typos are likely correct for that rendering, let me know if this is the case.

jbd:
I'm reading the documentation from https://singularityhub.github.io/sregistry/install

vsoch:
Ah good catch! Yes the approach I think is most sound is to have the full path link to the file.

vsoch:
Thanks for the fix!

vsoch:
Also, if you are interested, the Globus integration is under review #113 and definitely needs reviewers! @victorsndvg is on the queue but I think he's very busy.

victorsndvg:
Yep @vsoch , I'm very busy ... but excited by upcoming tests with globus! I will report as soon as possible ;)

PR #132, opened 2018-06-04
Fix/messages

vsoch:
@victorsndvg does this ok to resolve the remaining issues?

### DIFF
--- a/shub/apps/users/templates/users/_profile_integrations.html
+++ b/shub/apps/users/templates/users/_profile_integrations.html
@@ -15,7 +15,7 @@
             <a class="social-button btn-lg" 
                id="globus-login" 
                style="padding-left:60px; padding-right:30px"
-               href="{% url 'globus_login' %}?next=http://localhost{{ request.path }}">Connect Globus</a><hr>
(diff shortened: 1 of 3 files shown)


victorsndvg:
mmmm

Hi @vsoch, it seems that there is no `request` in some scopes. Maybe it has to be passed as an argument to some functions? or maybe I'm missing something?

Look at this lines:

https://github.com/singularityhub/sregistry/blob/e03a0b7ea85ec6d8639e3b5241927d826f70d5f7/shub/apps/users/views/teams.py#L224
https://github.com/singularityhub/sregistry/blob/e03a0b7ea85ec6d8639e3b5241927d826f70d5f7/shub/apps/users/views/teams.py#L240
https://github.com/singularityhub/sregistry/blob/e03a0b7ea85ec6d8639e3b5241927d826f70d5f7/shub/apps/users/views/teams.py#L243

vsoch:
Ah, this is because the first "fix" to add the request was wrong to begin with! Those underscore (private) functions don't return directly to views, and the messages being there was legacy from when they returned json response. The messages shouldn't be there because they don't return to a view, and requests shouldn't be defined / used in the functions at all.

victorsndvg:
Perfect, working! :+1:

vsoch:
Awesome!

PR #150, opened 2018-09-12
updating uwsgi to handle downloads

vsoch:
@victorsndvg would you care to look if this is the configuration that works better for the setup with nginx?


victorsndvg:
@vsoch , yes this is my current configuration. If I find any issue I will report it!

vsoch:
ok cool, I'm going to merge to make default, it's logical.

PR #159, opened 2018-10-11
Updating error to catch filename error also on OSError

### LINKED ISSUE #158: filename reveal should not show path on system

vsoch:
Currently, we catch the "file is too long" error with a DataError when the object is created. The error can be triggered earlier when we try to move the nginx upload to storage, so I've moved this into the try catch, and also am just printing the basename of the file so the user doesn't see where it's being stored. @victorsndvg this should address your concern and it would be great it you might test, and will close #158


victorsndvg:
Great, thanks!

Please, feel free to merge

PR #188, opened 2019-04-19
making sregistry quiet for warning

vsoch:
This PR can be used to test the custom handlers (note that I tried changing the function to also take an exception and then produced the error, having one argument did not). So currently this PR is just to quiet the sregistry client warnings (that seem to be confusing to the user, and aren't helpful)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM python:3.5.7
 ENV PYTHONUNBUFFERED 1
 ENV DEBIAN_FRONTEND noninteractive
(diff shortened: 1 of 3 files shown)


vsoch:
I also just found that the 500 handler is the only one that takes one argument (the rest have two) so I updated based on seeing this here https://github.com/django/django/blob/48493636d3994d1b80d2bf213283240521d368a0/django/urls/resolvers.py#L403

vsoch:
@sgabenov  you can test this branch to see if it fixes the warnings and 404Handler issue.

PR #200, opened 2019-06-12
Add library API endpoint

### LINKED ISSUE #190: Support Singularity 3 library:// protocol
Singularity 3.x introduced the new library:// (used e.g. for https://cloud.sylabs.io/library) protocol which makes the whole sregistry client obsolete (from what I checked the singularity executable supports all operations the sregistry client supports, not 100% sure though)

Features:
- push
- pull
- search
- authentication (via token)

It would be great if this was supported by the server too possibly replacing (maybe via config) the shub:// protocol (which is the predecessor of library://)

vsoch:
This PR will supplement the work to add builders, specifically adding a mirror of the library API to close #190.  We also:

 - calculate the sha256 sum as a task on image upload
 - add a much nicer schema view for restful apis
 - prepare to receive a hash from the Google Builder


vsoch:
This is good enough to add a basic pull. I want to work on this on the server, so merging.

PR #215, opened 2019-07-12
Minor fixes and improvements

tschoonj:
Ran into some problems as I was setting up the server, thought they may be helpful to others as well.

### DIFF
--- a/scripts/generate_cert.sh
+++ b/scripts/generate_cert.sh
@@ -14,7 +14,7 @@ sudo apt-get update
 sudo apt-get install python-certbot-nginx
 
 # Get certificates (might need sudo)
-sudo certbot certonly --nginx -d "${DOMAIN}" -d "www.${DOMAIN}"--email "${EMAIL}" --agree-tos --redirect
+certbot certonly --nginx -d "${DOMAIN}" -d "www.${DOMAIN}" --email "${EMAIL}" --agree-tos --redirect
 
 # The prompt is interactive, and will show the locations of certificates
 
--- a/scripts/generate_cert.sh
+++ b/scripts/generate_cert.sh
@@ -14,7 +14,7 @@ sudo apt-get update
 sudo apt-get install python-certbot-nginx
 
 # Get certificates (might need sudo)
-certbot certonly --nginx -d "${DOMAIN}" -d "www.${DOMAIN}" --email "${EMAIL}" --agree-tos --redirect
+sudo certbot certonly --nginx -d "${DOMAIN}" -d "www.${DOMAIN}"--email "${EMAIL}" --agree-tos --redirect
 
 # The prompt is interactive, and will show the locations of certificates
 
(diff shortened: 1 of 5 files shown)


scripts/generate_cert.sh
```diff
@@ -14,7 +14,7 @@ sudo apt-get update
 sudo apt-get install python-certbot-nginx
 
 # Get certificates (might need sudo)
-certbot certonly --nginx -d "${DOMAIN}" -d "www.${DOMAIN}" --email "${EMAIL}" --agree-tos --redirect
+sudo certbot certonly --nginx -d "${DOMAIN}" --email "${EMAIL}" --agree-tos --redirect
```
vsoch: This is actually wrong for most - if you don’t issue certs for www as well it leads to all kinds of trouble! Please leave as is, and if you want to add a note about cases when it’s not necessary that would be appropriate.

vsoch:
Also for the server we have a CHANGELOG.md and the version string is in the VERSION file.

scripts/generate_cert.sh
tschoonj: Good morning @vsoch!

This failed for me because I never created a record for `www.${DOMAIN}` in Route 53. If I would have, would this have led to the user being redirected to `https://${DOMAIN}` if he/she typed `https://www.${DOMAIN}` in the browser? Or would this have needed additional Nginx configuration? 

Sorry if this sounds dumb... am quite far outside of my comfort zone here...

vsoch: I think we should provide the instruction for how to do it, and then (as you did) the user can delete it, if not appropriate. I think for most professional sites, www is defined.  How about something like:

````bash
# Get certificates (might need sudo)
certbot certonly --nginx -d "${DOMAIN}" -d "www.${DOMAIN}" --email "${EMAIL}" --agree-tos --redirect
```
or if you haven't defined www in your DNS, you can remove that from the request:

````bash
certbot certonly --nginx -d "${DOMAIN}" --email "${EMAIL}" --agree-tos --redirect
````
Let's also leave the actual sudo out, since it says "may need sudo" and the user can add back in if needed.

File: scripts/generate_cert.sh [hunk 2]
```diff
@@ -14,7 +14,7 @@ sudo apt-get update
 sudo apt-get install python-certbot-nginx
 
 # Get certificates (might need sudo)
-certbot certonly --nginx -d "${DOMAIN}" -d "www.${DOMAIN}" --email "${EMAIL}" --agree-tos --redirect
+certbot certonly --nginx -d "${DOMAIN}" -d "www.${DOMAIN}"--email "${EMAIL}" --agree-tos --redirect
```
vsoch: looks like there is a space missing here!

VERSION
```diff
@@ -1 +1 @@
-1.1.01
+1.1.02
```
vsoch: And actually I'm an idiot - we don't need to bump the version if there aren't changes to the server. Sorry for missing this - I had just woken up :P

scripts/generate_cert.sh
tschoonj: Grr.... fixing!

vsoch:
Ready for merge?

tschoonj:
nope, gimme a minute

tschoonj:
Ready when you are...

tschoonj:
By the way: next week we are going to deploy a singularity registry instance on our K8S cluster... Have you ever tried doing that?

vsoch:
I personally haven't, but others have: https://github.com/sylabs/singularity-cri/issues/321

I don't have a bunch of servers to throw around and test, but I'd hugely like to support this! So whatever help you need, development work, etc., count me in.

tschoonj:
> I personally haven't, but others have: [sylabs/singularity-cri#321](https://github.com/sylabs/singularity-cri/issues/321)
> 
> I don't have a bunch of servers to throw around and test, but I'd hugely like to support this! So whatever help you need, development work, etc., count me in.

I will keep your kind offer in mind should the need arise for your assistance 👍

Issue #225, opened 2019-08-22
nginx ldap_auth not found


groucho64738:
Hi. I'm installing the sregistry with LDAP support and followed the directions, but I'm running into an issue where it seems like nginx can't find ldap_auth:

uwsgi_1      | [pid: 37|app: 0|req: 1/1] 10.251.16.20 () {38 vars in 592 bytes} [Thu Aug 22 09:52:40 2019] GET / => generated 8220 bytes in 690 msecs (HTTP/1.1 200) 4 headers in 124 bytes (1 switches on core 0)
uwsgi_1      | [pid: 38|app: 0|req: 1/2] 10.251.16.20 () {40 vars in 650 bytes} [Thu Aug 22 09:52:42 2019] GET /login/ => generated 5720 bytes in 660 msecs (HTTP/1.1 200) 4 headers in 124 bytes (1 switches on core 0)
uwsgi_1      | Not Found: /ldap_auth/login/
uwsgi_1      | [pid: 38|app: 0|req: 2/3] 10.251.16.20 () {40 vars in 759 bytes} [Thu Aug 22 09:52:44 2019] GET /ldap_auth/login/?next=http://myhost/login/ => generated 652 bytes in 26 msecs (HTTP/1.1 404) 3 headers in 116 bytes (1 switches on core 2)

I made sure in the Dockerfile and on the build line that I'm enabling the LDAP_AUTH plugin, and I see that django-ldap-auth and stuff is being added during the container builds.

Anyone have a hint on what to look for next? Thanks for your help.

vsoch:
Did you follow the instructions and uncomment ldap_auth from plugins installed? The second bullet here: https://singularityhub.github.io/sregistry/docs/plugins/ldap

https://github.com/singularityhub/sregistry/blob/master/shub/settings/config.py#L133

The message is telling us that django didn't find the urls associated with the login, so I'm guessing that is still commented.

groucho64738:
I have this in the shub/settings/config.py

PLUGINS_ENABLED = [
    'ldap_auth',
#    'google_build'
#    'pam_auth',
#    'globus',
#    'saml_auth'
]

vsoch:
Did you restart the containers after uncommenting?

groucho64738:
yes I did a docker-compose down and up again. Is there something inside of the uwsgi container itself when its restarted that I can look at to verify?

vsoch:
[...]

PR #226, opened 2019-08-22
Fixing up issue with login url for ldap not being accessible

vsoch:
This is a quick fix to correct an issue with the user profile view (that accepts a username) not having the correct prefix to allow for access to (likely many) other urls, including ldap. I also fixed a bug with the view to delete a user account - I thought I fixed it before but must have changed the wrong file.

Signed-off-by: Vanessa Sochat <vsochat@stanford.edu>


vsoch:
This will close #225

Issue #273, opened 2020-02-24
pushed container with tag "latest" does not update


RonaldEnsing:
The status of a new container pushed to sregsitry is 'frozen', meaning that it cannot be overwritten. I would like to have new containers pushed to sregistry to have the status 'open' by default, so that I can overwrite containers without having to login the web interface to change the status manually, e.g. when I push containers from a CI. Is this option available? If so, how do I configure this?

edit: Maybe I should open a separate issue about this, but it also looks like the container does not get overwritten after setting status to 'open'. If I push another container (different content), the time stamp is modified but the container hash is not.

vsoch:
Hmm, I'm not sure where this is happening, because for the create command, frozen is set to False (see here) https://github.com/singularityhub/sregistry/blob/79ab935a2e51acfd7b22dca8d72382a8aabb318c/shub/apps/library/views/images.py#L226 and the default is also False https://github.com/singularityhub/sregistry/blob/master/shub/apps/main/models/containers.py#L79. Can you trace the operation you are doing - is it possibly the case you are hitting a container that already exists, or that there is a create somewhere we are overlooking?

RonaldEnsing:
Not sure what happened. It could be that the container existed before. I am not able to reproduce, even with a container that existed before. So I'll close this issue.

RonaldEnsing:
I was able to reproduce the issue after all. Although it turns our slightly different compared to what I stated before. The status is indeed 'Open' by default. However, when I push a container in order to overwrite an existing container, the time stamp in the web interface is modified, but the original container is not overwritten.

vsoch:
okay so just to clarify your situation:

 - you have a collection with a container that is not frozen
 - you push a new container with the same name _and_ tag (what is the tag?) and you expect the binary to be over-written
 - you see a timestamp update in the interface, but the container itself is the previous one (validated I assume with a pull and then checking some variable / version inside?)

RonaldEnsing:
> okay so just to clarify your situation:
> 
>     you have a collection with a container that is not frozen
> 
Yes
>     you push a new container with the same name _and_ tag (what is the tag?) and you expect the binary to be over-written

Yes. I simply used some custom tag ('latest').

>     you see a timestamp update in the interface, but the container itself is the previous one (validated I assume with a pull and then checking some variable / version inside?)
>

Yes. Although I did not use a `singularity pull`. I downloaded the container from the web interface. I don't expect that would make a difference though.
 
> 
[...]

PR #279, opened 2020-03-05
Update to container push to honor new binary

vsoch:
This pull request will address #273 - specifically, we had a typo in the final tag function where, given that a container was not frozen, we would want to keep the newly uploaded one and clean up the previous one. In addition, the django signal wasn't firing to clean up any image files that were no longer used.

To test this pull request:

 - keep in mind that since the docker-compose has a parameter added to postgres (container df) you should not test this on a production running instance, because there is a chance of the container being re-created and losing the database.
 - if you are testing locally you will need to follow instructions to compile a singularity version that changes a line in remote.go to http from https (details [here](https://singularityhub.github.io/sregistry/docs/client)) and also there are detail for how to define and login to the remote. Don't forget to generate your secrets.py file with a secret_key!

## Things to test
- [ ] creating a new collection and pushing a container
- [ ] push of a container with a new tag
- [ ] push of a container with an existing tag
- [ ] push of the same container with a different tag

I tested these points, and I'd like one more tester for a sanity check. @RonaldEnsing I'm hoping since the issue is important to you that you might be able to - and please there is no rush! Take whatever time that you need. 


RonaldEnsing:
Yes, this works. Thanks a lot! I think this is good to merge.

Issue #290, opened 2020-03-26
Is DATA_UPLOAD_MAX_MEMORY_SIZE actually used?


Aneoshun:
Hi !

I have been using SRegistry on my Synology NAS without issue for a couple of months now, and that's really great. 
However, I am starting to manipulate significantly larger containers (about 3GB, while so far I used 700MB). My NAS has only 2GB of RAM (yeah... I should definitely increase it, but in lockdown condition, it's, for now, impossible), and the push of my large containers fails with a 403 error (in addition to perturbing the other services running on the NAS). The 403 error received from the client, but the SRegistry logs show things like this: 
```
[error] 6#6: *3392 upstream timed out (110: Operation timed out) while reading response header from upstream, client: 172.17.0.1, server: localhost, request: "PUT /v2/push/imagefile/261/SHA_REDACTED HTTP/1.1", upstream: "uwsgi://172.17.0.4:3031", host: "REDACTED" 
```

In `config.py`, I set `DATA_UPLOAD_MAX_MEMORY_SIZE = 500`, but it does not seem to help. 

When I use the search bar in GitHub, DATA_UPLOAD_MAX_MEMORY_SIZE seems to appear only once on the repo (in the config.py). So, I am wondering if this setting is actually used and whether there is a way to improve the situation on my side?

Note: copying the same containers on the same NAS via NFS works without issue, so it is not a network capability issue. 

I will be very glad if you have a suggestion on this. 

Thanks a lot in advance. 

Best regards and stay safe!

vsoch:
hey @Aneoshun ! I've had this issue as well, and it's always when I'm low on memory, and I can tell because my computer starts to freeze up during upload of some large container. It shows up in different ways (sometimes I see an EOF) but the fix is always clearing up memory. I don't have a quick fix solution because singularity handles the logic of the upload, but generally I'd recommend deploying the registry on a host that has maybe at least 8GB. The computer I am on now (that has 16GB) can even have issues with large container pushes when memory is running low.

[...]

PR #291, opened 2020-03-26
Update/max memory size

vsoch:
Currently, this value is set to None (and incorrectly says to use GB and not bytes) This will define the value to be it's default, add links to the page, and also add `FILE_UPLOAD_MAX_MEMORY_SIZE`. This will hopefully address #290


vsoch:
This issue was fixed by adding the minio backend, so closing here.

Issue #301, opened 2020-05-25
Scheduler and Worker containers stop after a minio.error.AccessDenied


Aneoshun:
Hi @vsoch, 

I have noticed that after a couple of hours, the Scheduler and Worker containers crash, both throwing this exception: 
`minio.error.AccessDenied: AccessDenied: message: Access Denied`

However, uploading and downloading seem to still work. 

Do you have an idea of what could be causing this issue? Maybe I missed something in the configuration? 

Thanks, 

Here are the complete traceback:
Scheduler: 
```Traceback (most recent call last):

  File "/code/manage.py", line 9, in <module>

    execute_from_command_line(sys.argv)

  File "/usr/local/lib/python3.5/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line

    utility.execute()

  File "/usr/local/lib/python3.5/site-packages/django/core/management/__init__.py", line 375, in execute

    self.fetch_command(subcommand).run_from_argv(self.argv)

  File "/usr/local/lib/python3.5/site-packages/django/core/management/base.py", line 323, in run_from_argv

    self.execute(*args, **cmd_options)

  File "/usr/local/lib/python3.5/site-packages/django/core/management/base.py", line 361, in execute

    self.check()

  File "/usr/local/lib/python3.5/site-packages/django/core/management/base.py", line 390, in check

    include_deployment_checks=include_deployment_checks,

  File "/usr/local/lib/python3.5/site-packages/django/core/management/base.py", line 377, in _run_checks

    return checks.run_checks(**kwargs)

  File "/usr/local/lib/python3.5/site-packages/django/core/checks/registry.py", line 72, in run_checks

    new_errors = check(app_configs=app_configs)

  File "/usr/local/lib/python3.5/site-packages/django/core/checks/urls.py", line 40, in check_url_namespaces_unique

    all_namespaces = _load_all_namespaces(resolver)

  File "/usr/local/lib/python3.5/site-packages/django/core/checks/urls.py", line 57, in _load_all_namespaces

    url_patterns = getattr(resolver, 'url_patterns', [])

  File "/usr/local/lib/python3.5/site-packages/django/utils/functional.py", line 80, in __get__

    res = instance.__dict__[self.name] = self.func(instance)

  File "/usr/local/lib/python3.5/site-packages/django/urls/resolvers.py", line 584, in url_patterns

    patterns = getattr(self.urlconf_module, "urlpatterns", self.urlconf_module)

[...]

PR #302, opened 2020-05-25
Adding minio env file to scheduler and worker

vsoch:
This will address #301 - it's very likely that the worker and scheduler do the same imports, and this will result in an error (and then exit nonzero of the container) if we don't provide the same environment variables. My quick first test fix is to provide the same credential file. 


Aneoshun:
Hi @vsoch, 

This solves #301 on my side. After 16h of execution, all the containers are still alive. 
Thanks!

vsoch:
Phew! Really good to hear. Let’s merge right away then!

PR #332, opened 2020-12-21
enforce lowercase of username when created

vsoch:
This is a test PR to enforce lowercase of a new username, and we do this because the scs library client now requires it: https://github.com/sylabs/scs-library-client/blob/31250b63795e4a20afd8f4f00b2ad3b34ec188e0/client/util.go#L28


vsoch:
I have to merge and release this blindly, as the user that requested the change doesn't seem to want to help.

PR #339, opened 2021-01-26
(maint) update docker-compose.yml to version 3

vladdoster:
Update docker-compose.yml from version 1 to 3. Not as
straight forward as it seemed.

### DIFF
--- a/docker-compose-https.yml
+++ b/docker-compose-https.yml
@@ -1,72 +1,76 @@
 version: "3.7"
 
-
 services:
 
   db:
     image: postgres:latest
     environment:
       - POSTGRES_HOST_AUTH_METHOD=trust
-  
+
   uwsgi:
-    restart: on-failure
     image: quay.io/vanessa/sregistry
     env_file:
       - ./.minio-env
+    restart: on-failure
     volumes:     
       - .:/code
       - ./images:/var/www/images
       - ./static:/var/www/static
-  
+
   nginx:
-    restart: on-failure
     image: quay.io/vanessa/sregistry_nginx
+    restart: on-failure
     ports:
       - "80:80"
+      - "443:443"
     volumes:
-      - ./images:/var/www/images
-      - ./nginx.conf:/etc/nginx/conf.d/default.conf:ro
-      - ./static:/var/www/static
+      - ./https.nginx.conf:/etc/nginx/conf.d/default.conf:ro
       - ./uwsgi_params.par:/etc/nginx/uwsgi_params.par:ro
-      - .:/code
-  
-  redis:
-    restart: on-failure
-    image: redis:latest
-  
-  scheduler:
-    image: quay.io/vanessa/sregistry 
-    volumes:
-      - .:/code
+      - /etc/ssl/certs:/etc/ssl/certs:ro
+      - /etc/ssl/private:/etc/ssl/private:ro
       - ./images:/var/www/images
       - ./static:/var/www/static
-    env_file:
-      - ./.minio-env
-    command: python /code/manage.py rqscheduler
-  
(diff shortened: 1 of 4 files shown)


vsoch:
Oh hmm, interesting. To play devil's advocate, I use a fairly recent version of docker-compose, and haven't had issue running the config files. If we update to a newer one, then folks with older versions will probably lose functionality. What are the reasons to do this, and why is it a better approach than the current?

vladdoster:
https://docs.docker.com/compose/compose-file/compose-versioning/#version-3

You gain new syntax and allows us to take advantage of new docker engine features.

We weren't able to use docker swarm or Kubernetes with version 1.

vsoch:
Oh, neat! That definitely is an improvement. If you've tested this and it works as expected, we should be good! I would ask if you could also update the https example? https://github.com/singularityhub/sregistry/tree/master/https

vsoch:
And to have a quick discussion about releases - merging into master will actually deploy a container with the current release in the VERSION file, so I think since this is a fairly big change we'd want to also bump that to `1.1.31`, otherwise it will build over the previous container version.

vladdoster:
Will wrap this up this weekend. Apologize for the delay

vsoch:
You never have to apologize for helping to make the project better - I'm here and ready when you are!

docker-compose-https.yml
```diff
@@ -0,0 +1,76 @@
+version: "3.7"
```
vsoch, line 1: This file should be in the https folder, so `https/docker-compose.yml` (so we should see a diff here!)

https/docker-compose.yml
```diff
@@ -1,82 +1,76 @@
-db:
-  image: postgres
-  environment:
-   - POSTGRES_HOST_AUTH_METHOD=trust
+version: "3.7"
```
vsoch, line 1: Oh I see you did edit this file - any reason to have two versions?

https/docker-compose.yml
```diff
@@ -1,82 +1,76 @@
...
-scheduler:
-  image: quay.io/vanessa/sregistry
-  command: python /code/manage.py rqscheduler
-  volumes:
-    - .:/code
-  volumes_from:
-    - uwsgi
-  env_file:
-    - ./.minio-env
-  links:
```
vsoch, line 53: So the newer version doesn't need to define links? And `volumes_from` no longer exists so each needs to have all volumes listed explicitly?

docker-compose-https.yml
```diff
@@ -0,0 +1,76 @@
...
+      - uwsgi
+    env_file:
+      - ./.minio-env
+    volumes:
+      - .:/code
+      - ./images:/var/www/images
+      - ./static:/var/www/static
+    command: python /code/manage.py rqworker default
+
+volumes:
```
vsoch, line 73: Where are these volumes referenced? Typically volumes here refer to volumes created (docker) and then you would reference them in a volume definition. I think we just want the containers to bind to the host, so I'm not sure they are needed.

https/docker-compose.yml
```diff
@@ -1,82 +1,76 @@
...
+  db:
+    image: postgres:latest
+    environment:
+      - POSTGRES_HOST_AUTH_METHOD=trust
 
-redis:
-  restart: always
-  image: redis:latest
+  uwsgi:
+    restart: always
```
vsoch, line 11: Is there reason that these restarts are `always` but the other ones are changed to `on-failure`?

cpeel:
https://github.com/singularityhub/sregistry/pull/405 updated `docker-compose.yml` to use the new format and I think this can be closed.

PR #341, opened 2021-01-30
(maint) update prepare_instance.sh

vladdoster:
- quote variables to avoid globbing

- cd error handling
> cd can fail for a variety of reasons: misspelled paths, missing directories,
> missing permissions, broken symlinks and more. If/when it does, the script will
> keep going and do all its operations in the wrong directory. This can be messy,
> especially if the operations involve creating or deleting a lot of files.

### DIFF
--- a/scripts/prepare_instance.sh
+++ b/scripts/prepare_instance.sh
@@ -58,15 +58,15 @@ sudo docker run hello-world
 # sudo groupadd docker
 
 #make sure to add all users that will maintain / use the registry
-sudo usermod -aG docker $USER
+sudo usermod -aG docker "$USER"
 
 # Docker-compose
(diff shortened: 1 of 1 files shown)


vsoch:
The issue with these changes are that if the cd does fail, there is no message to the user (it's silent). And since we aren't deleting files, there really isn't any risk to doing it - we just are building a docker container, and likely if the user doesn't cd they will see a message that there is no Dockerfile in the context. This script is also not really intended to be run non-interactively, at least I mostly did this manually. So I think the changes to add quotes are good, but I'm not sure we need `cd X || exit` - it seems like a bit much, and could actually be confusing. Would it make sense to instead check that the install path exists before cding there?

Issue #344, opened 2021-02-18
How to create a new collection or a new container using the API with curl?


pini-gh:
I need a way to create new private / public collections / containers via the command line.
Thanks in advance.

vsoch:
I had originally removed this because I was concerned about some kind of spammy operation, but with authentication it's reasonable. Would you care to open a PR to do it? I just started a new job so I won't have time for at least a few weeks.

pini-gh:
Well, I'm all for doing it, but I'd need a few hints :/ If this was removed, maybe there is an old commit I could start with?

vsoch:
The previous implementation used the (now mostly) deprecated original push/pull container API, which is replaced with the library API. The first thing I'd do is check with sylabs if their library API has support for this kind of creation, and if yes, add it to the library API. If not, then it can go as a separate API function that has the same authentication check, and some ratelimiting.

pini-gh:
Yes, their API do have support for it. Actually I need to set up a separate repository from an existing Sylabs installation, and I think I just miss these two features to be able to go with sregistry.

vsoch:
Ah, great! So just reproduce those exact endpoints. You'll likely want to add a new file for collections in this folder https://github.com/singularityhub/sregistry/tree/master/shub/apps/library/views and you can copy the style of functions from images.py, for example (to have the correct ratelimiting and authentication checks). What I did to design this API was basically trace / look at the Sylabs SCS Client, so if they have functions for a new collection it should be somewhere in https://github.com/sylabs/scs-library-client. You can ask them on slack if it's not easy to find (I'm not familiar with this endpoint via Singularity).

pini-gh:
Is that what you're talking about?
https://github.com/sylabs/scs-library-client/blob/82d22e604229c73a27d339106ebcb2c6e137e03f/client/api.go#L78-L112

vsoch:
yes that's perfect! In that function you can see what the data is expected to look like, and the response. When I was developing the original library API, what I found helpful was to compile a version of singularity with print statements so I could see what the responses were expected to look like (and do this first against the Sylabs cloud for a working operation). Then you basically just duplicate that for sregistry.

pini-gh:
With the sregistry data model, is it possible to have several collections with the same name, provided they are associated with different entities? It is possible with Scilabs, but here it seems I can search a collection by its name only, and not by its name + entity.

vsoch:
[...]

PR #345, opened 2021-02-19
Add API endpoint to create a collection

pini-gh:
API path: /v1/collections
Method: POST
Parameters:
* 'entity': owner's entity id (= user.id)
* 'name': nmae for the new collection
* 'private' (optional): boolean; default to get_privacy_default()

Partial implementation for singularityhub/sregistry#344


shub/apps/library/views/images.py
```diff
@@ -551,6 +554,60 @@ def get(self, request):
...
+
+        print("POST CollectionsView")
+        if not validate_token(request):
+            print("Token not valid")
+            return Response(status=403)
+
+        # body should have {'entity': entity_id, 'name': new_collection_name, 'private': bool}
+        # 'private' is optional and defaults to Collection.private default
+        # value set with get_privacy_default()
+        # {"entity": "42", "name": "my_collection", "private": true}
```
vsoch, line 579: Although sylabs might use an entity ID, the user making the request isn't easily going to know their. So let's change this (and the subsequent query) to look for the user's `user.username`.

shub/apps/library/views/images.py
```diff
@@ -551,6 +554,60 @@ def get(self, request):
...
+        body = json.loads(request.body.decode("utf-8"))
+        if not ("entity" in body and "name" in body):
+            print("Invalid payload")
+            return Response(status=400)
+
+        # does a collection with the same name exist already?
+        if not get_collection(body["name"], False) is None:
```
vsoch, line 586: You don't have to say `is None` you can just say `if not X`.

shub/apps/library/views/images.py
```diff
@@ -551,6 +554,60 @@ def get(self, request):
...
+            print("A collection named '{0}' exists already!".format(body["name"]))
+            return Response(status=403)
+
+        # check user permission
```
vsoch, line 590: We should be validating if the user has permission to do this upfront, before giving them information about existing, missing params, etc. They should be denied right away of this is off.

shub/apps/library/views/images.py
```diff
@@ -551,6 +554,60 @@ def get(self, request):
...
+        token = get_token(request)
+        if str(token.user.id) != body["entity"]:
+            print("Permission denied {0} {1}".format(token.user.id, body["entity"]))
+            return Response(status=403)
+
+        # create the collection
+        name = format_collection_name(body["name"])
+        collection = Collection(name=name, secret=str(uuid.uuid4()))
```
vsoch, line 598: Do you not want to do a proper `Create` here (which would handle that first save?) And could this fail for any reason and throw an exception?

vsoch (commented):
This is a great start @pini-gh ! I had a few comments about ordering and syntax, but otherwise the endpoint looks good to test. Could you please add to the documentation how to go about this? I think you would want to write up the example at the end of this Singularity section https://github.com/singularityhub/sregistry/blob/master/docs/_docs/client.md#singularity and you can also add the equivalent operations for curl there.

shub/apps/library/views/images.py
vsoch: But this commenting is great - by all means please leave the commenting for the developer to see!

vsoch: And if it's the case that Singularity uses another endpoint to get the ID first based on the user provided usrename and then issues this request to create the collection, then it would be ok to leave as is. I think we'd want to note in the docs (with the curl example perhaps?) How to look up the user id.

shub/apps/library/views/images.py
pini-gh: Well, I'm not familiar at all with the django framework and didn't find such a method in the source tree. How is this `Create` supposed to be used?

shub/apps/library/views/images.py
pini-gh: Moving it one block up. Can't do this before checking the parameters, because the entity id comes from these.

shub/apps/library/views/images.py
pini-gh: OK.

shub/apps/library/views/images.py
vsoch: If you think something might exist, you do:

```python
instance, created = Model.objects.get_or_create(name=...)
```
But since you know this doesn't exist, you would just do:
```python
instance = Model.objects.Create(name=...)
```
It saves you the step of needing to do `instance.save()`. For the way you did it, typically that's done if you need to create the instance (without saving) because it's missing some required field.

vsoch: There is good docs starting here https://docs.djangoproject.com/en/3.1/ref/models/querysets/#create

shub/apps/library/views/images.py
vsoch: Correct - the important thing is to move it above the block that tells them that something exists (or not). An unathenticated user could basically discover a lot of containers that way.

shub/apps/library/views/images.py
pini-gh: Not sure what you mean. My aim is to replicate the Sylabs endpoint actually. It wouldn't make sense to change its signature.

pini-gh: In the sregistry data model the Sylabs entity id corresponds to the user id.

vsoch: So what endpoint is the Sylabs client using to retrieve (or lookup) a user to get the id in the first place? Is this endpoint already supported in sregistry?

shub/apps/library/views/images.py
pini-gh: How about this?
```python
        name = format_collection_name(body["name"])
        try:
            collection = Collection.objects.get(name=name)
            print("A collection named '{0}' exists already!".format(name))
            return Response(status=403)
        except Collection.DoesNotExist:
            pass

        collection.secret = str(uuid.uuid4())
        collection.owners.add(token.user)

        # set privacy if present
        if "private" in body:
            collection.private = body["private"]

        collection.save()
```

vsoch: Sorry, I think you were working on the wrong thing here? This functionality is already handled above by `get_collection`. The Create that we want to handle is below that, you currently have:

```python
collection = Collection(name=name, secret=str(uuid.uuid4()))
```
And as I mentioned above, you'd want to use Create

```python
collection = Collection.objects.create(name=name, secret=str(uuid.uuid4()))
```
You've already established that it does not exist with `get_collection` before that.

pini-gh: I've grouped both topics (check then create) into this snippet.

shub/apps/library/views/images.py
pini-gh: I don't know which endpoint the Sylabs client uses, because I use CLI (bash + curl) with *this* endpoint  the Sylabs library, and I need the sregistry's endpoint to be the same.

vsoch: This is why you should test with the Singularity client. If an endpoint is needed to retrieve the user first, we'll add it here.

vsoch:
I just pushed some suggestions:

 - we do a get_or_create to either get the collection (and return an error if it exists) or create it if it does not.
 - I added error messages to the responses, as it looks like the sylabs json parser expects an error to be formatted as such (and likely will print it to the calling user) https://github.com/sylabs/json-resp/blob/master/json_response.go#L16

pini-gh:
> Could you please add to the documentation how to go about this? I think you would want to write up the example at the end of this Singularity section https://github.com/singularityhub/sregistry/blob/master/docs/_docs/client.md#singularity and you can also add the equivalent operations for curl there.

Well, I don't see any documentation there about the REST API. Then it seems weird to write about this endpoint only.
BTW thanks to the docstrings, it is automagically documented via the `/api/` path.

vsoch:
@pini-gh that is the client documentation that shows the user how to interact with different endpoints (push, pull, etc.) We would want a section for creating a collection, and since you are using Curl it makes sense to have Singularity and curl as well.

shub/apps/library/views/images.py
pini-gh: Maybe you could explain what you mean by "retrieve the user"? Would it be "retrieve the entity id from its name"? In that case this can be done with the `/v1/entities/{user}` endpoint. It returns an entity object which features an `id` attribute.

BTW I am just a user of our Sylabs instance. I don't have any admin access and won't be granted one. Then I can't trace which endpoints are involved when the Sylabs client interact with the library.

pini-gh (commented):
#Would this match what you have in mind?
# Curl
Like with the Sylabs Library API, it is possible to interract with the Singularity Registry Server using Curl. You can browse the API schema via the `/api` path of your server.

## Authentication
Authentication is done presenting the token in an `Authorization` header:
```
$ curl -s -H 'Authorization: Bearer <token>' https://shub.example.com/<api_endpoint>
```
## Create a collection
As of version `1.1.32` it is possible to create a new collection via the API. It requires authentication.

First retrieve the numeric `id` associated with your username with a GET request to the endpoint `/v1/entities/<username>`.

Then issue a POST request to the endpoint `/v1/collections` and this json payload:
```
{
  "entity": "<user_id>"
  "name": "<new_collection_name>"
  "private": true|false
}
```

The `private` key is optional. If not provided, it defaults to the servers's configured defaut for collection creation.
#

vsoch:
Yes that looks great!

So basically, if the Sylabs library client has a function to create a collection, we'd want to understand the context this is used from Singularity, so we can test that out with Singularity Registry server. Since the Sylabs API will identify the user with a token, and also since there is the entity id in the request, I figure there is a mechanism that first identifies the user, and then prepares the request with the user ID. In a nutshell, we want to be able to test sregistry with the singularity client, and the first thing is to figure out how and when that call is made, and see if the current implementation can support it.

I can look at this if it's unclear, but I won't be able to for a little bit. Your docs so far (with curl) look great, btw! :)

pini-gh:
> In a nutshell, we want to be able to test sregistry with the singularity client, and the first thing is to figure out how and when that call is made, and see if the current implementation can support it.

As far as I know the singularity client has no feature to create collections. I still don't understand what you want to test on that matter.

How would it be a blocker regarding this PR?

vsoch:
I can show you. As I suspected, the client gets / creates the entity here:

https://github.com/sylabs/scs-library-client/blob/82d22e604229c73a27d339106ebcb2c6e137e03f/client/push.go#L132

Given that a collection doesn't exist, this data is then passed to the function here:

https://github.com/sylabs/scs-library-client/blob/82d22e604229c73a27d339106ebcb2c6e137e03f/client/push.go#L146

So to test this with Singularity, after following the instructions to [compile it](https://singularityhub.github.io/sregistry/docs/client) to allow http, you *should* be able to create a container and collection that does not exist. That is how you test it.

pini-gh:
OK, I understand now.

Just tested a `singularity push` to a non existent collection, and it works now! Looking at the uwsgi container logs I see a successful POST to the new endpoint:
```
[pid: 21|app: 0|req: 2/3] 172.18.0.4 () {44 vars in 614 bytes} [Sat Feb 20 20:10:07 2021] GET /v1/entities/pini => generated 317 bytes in 257 msecs (HTTP/1.1 200) 5 headers in 142 bytes (1 switches on core 1)
GET GetNamedCollectionView
Not Found: /v1/collections/pini/macollection
[pid: 21|app: 0|req: 3/4] 172.18.0.4 () {44 vars in 646 bytes} [Sat Feb 20 20:10:08 2021] GET /v1/collections/pini/macollection => generated 0 bytes in 63 msecs (HTTP/1.1 404) 4 headers in 115 bytes (1 switches on core 2)
POST CollectionsView
[pid: 21|app: 0|req: 4/5] 172.18.0.4 () {46 vars in 640 bytes} [Sat Feb 20 20:10:08 2021] POST /v1/collections => generated 350 bytes in 193 msecs (HTTP/1.1 200) 5 headers in 148 bytes (1 switches on core 3)
GET GetNamedContainerView
[pid: 21|app: 0|req: 5/6] 172.18.0.4 () {44 vars in 676 bytes} [Sat Feb 20 20:10:08 2021] GET /v1/containers/pini/macollection/scibian10-poste => generated 542 bytes in 131 msecs (HTTP/1.1 200) 5 headers in 142 bytes (1 switches on core 0)
GET PushNamedContainerView
[pid: 13|app: 0|req: 1/7] 172.18.0.4 () {44 vars in 833 bytes} [Sat Feb 20 20:10:08 2021] GET /v1/images/pini/macollection/scibian10-poste:sha256.3ad816fbe662e34f799e4b17cc10a5160293e88961f158d99388a8eccbf8539a?arch=amd64 => generated 560 bytes in 3525 msecs (HTTP/1.1 200) 5 headers in 142 bytes (1 switches on core 0)
[pid: 21|app: 0|req: 6/8] 172.18.0.4 () {44 vars in 596 bytes} [Sat Feb 20 20:10:12 2021] GET /version => generated 58 bytes in 11 msecs (HTTP/1.1 200) 5 headers in 141 bytes (1 switches on core 1)
POST RequestPushImageFileView
[pid: 17|app: 0|req: 2/9] 172.18.0.4 () {46 vars in 638 bytes} [Sat Feb 20 20:10:12 2021] POST /v2/imagefile/3 => generated 416 bytes in 143 msecs (HTTP/1.1 200) 5 headers in 137 bytes (1 switches on core 2)
```

vsoch:
Yes that’s perfect! So all you need to do is add a line in the make collection section to indicate that the create collection endpoint is used during a push in the case that the collection does not exist. I should then be able to make some time tomorrow to test this all out, approve and merge!

pini-gh:
Adding this at the end of the "Push logic" section's bulleted list:
>  - If you push an image to a non existing collection, the collection will be created first, then the image will be added (version `1.1.32`).

Adding this at the end of the "Create a collection" section:
> In case of a `singularity push` to a non existing collection, the client triggers the collection creation first, using this endpoint, then push the image.

vsoch:
ok! Just added a little more verbosity to the docs and I tested the curl, works like a charm! Take a look over my changes, if you approve, I'll merge first thing tomorrow.

pini-gh:
Seems all good. Thanks!

vsoch (approved):
Really great work on this @pini-gh ! It's been a pleasure.

PR #357, opened 2021-04-03
(maint) Dockerfile use `COPY` instead of `ADD`

vladdoster:
- Use `COPY` because it doesn't take advantage of the `ADD` auto `.tar` extraction feature.

Taken from Docker's [Best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy) documentation.


vsoch (approved):
This is a good call - I agree 100%.

PR #358, opened 2021-04-03
(maint) fix anti-patterns and improve performance

vladdoster:
- refactor unnecessary `else` / `elif` when `if` block has a `continue` statement
- refactor unnecessary `else` / `elif` when `if` block has a `raise` statement
- refactor unnecessary `else` / `elif` when `if` block has a `return` statement
- remove duplicate dictionary key(s)
- remove unnecessary return statement
- remove unnecessary use of comprehension
- use `_` to denote var is unused

- change methods not using its bound instance to staticmethods

    The method doesn't use its bound instance. Decorate this method with
    @staticmethod decorator, so that Python does not have to instantiate
    a bound method for every instance of this class thereby saving
    memory and computation.

- remove implicit `object` from the base class

    The class is inheriting from object, which is implicit under Python
    3 , hence can be safely removed from bases.

- use literal syntax instead of function calls to create data structure

    This is because here, the name dict must be looked up in the global
    scope in case it has been rebound. The same goes for the other two
    types list() and tuple().

    ```shell In [1]: timeit.timeit(stmt="dict()", number=100000000)
    Out[1]: 9.560388602000103

    In [2]: timeit.timeit(stmt="{}", number=100000000) Out[2]:
    1.685333584000091

    In [3]: timeit.timeit(stmt="tuple()", number=100000000) Out[3]:
    4.509182139000131

    In [4]: timeit.timeit(stmt="()", number=100000000) Out[4]:
    0.5455615430000762

    In [5]: timeit.timeit(stmt="list()", number=100000000) Out[5]:
    7.356000728000254

    In [6]: timeit.timeit(stmt="[]", number=100000000) Out[6]:
    1.573771306999788 ```

- fix membership test

    tests for membership should use the form `x not in the_list` rather
    than `not x in the_list`. the former example is simply more
    readable.

- use set comprehension

    although there is nothing syntactically wrong with this code, it is
    hard to read and can be simplified to a set comprehension. using set
    comprehension is more performant since there is no need to create
    another transient list.

- update singleton comparison

    comparisons to the singleton objects, like true, false, and none,
    should be done with identity, not equality. use “is” or “is not”.
    identity checks are faster than equality checks.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,11 +41,11 @@ RUN apt-get update && apt-get install -y \
     build-essential
 
 # Install Python requirements out of /tmp so not triggered if other contents of /code change
-ADD requirements.txt /tmp/requirements.txt
+COPY requirements.txt /tmp/requirements.txt
 RUN pip install --upgrade pip
 RUN pip install -r /tmp/requirements.txt
 
-ADD . /code/
+COPY . /code/
 
 ################################################################################
 # PLUGINS
--- a/shub/apps/api/urls/collections.py
+++ b/shub/apps/api/urls/collections.py
@@ -68,8 +68,7 @@ class CollectionViewSet(viewsets.ReadOnlyModelViewSet):
 class CollectionDetailByName(APIView):
     """Retrieve a collection instance based on it's name"""
 
-    @staticmethod
-    def get_object(collection_name):
+    def get_object(self, collection_name):
         try:
             collection = Collection.objects.get(name=collection_name.lower())
         except Collection.DoesNotExist:
@@ -97,8 +96,7 @@ class CollectionSearch(APIView):
     a general search to look across all fields for one term
     """
 
-    @staticmethod
-    def get_object(query):
+    def get_object(self, query):
         collections = collection_query(query.lower())
         return collections
 
--- a/shub/apps/api/urls/collections.py
+++ b/shub/apps/api/urls/collections.py
@@ -28,7 +28,8 @@ class CollectionSerializer(serializers.HyperlinkedModelSerializer):
 
     containers = serializers.SerializerMethodField("list_containers")
 
-    def list_containers(self, collection):
+    @staticmethod
+    def list_containers(collection):
         container_list = []
         for c in collection.containers.all():
             entry = {
@@ -67,7 +68,8 @@ class CollectionViewSet(viewsets.ReadOnlyModelViewSet):
 class CollectionDetailByName(APIView):
     """Retrieve a collection instance based on it's name"""
 
-    def get_object(self, collection_name):
+    @staticmethod
+    def get_object(collection_name):
         try:
             collection = Collection.objects.get(name=collection_name.lower())
         except Collection.DoesNotExist:
@@ -95,7 +97,8 @@ class CollectionSearch(APIView):
     a general search to look across all fields for one term
     """
 
-    def get_object(self, query):
+    @staticmethod
+    def get_object(query):
         collections = collection_query(query.lower())
         return collections
 
--- a/shub/apps/api/urls/containers.py
+++ b/shub/apps/api/urls/containers.py
@@ -36,20 +36,17 @@ class SingleContainerSerializer(serializers.ModelSerializer):
     image = serializers.SerializerMethodField("get_download_url")
     metadata = serializers.SerializerMethodField("get_cleaned_metadata")
 
-    @staticmethod
-    def collection_name(container):
+    def collection_name(self, container):
         return container.collection.name
 
-    @staticmethod
-    def get_cleaned_metadata(container):
+    def get_cleaned_metadata(self, container):
         metadata = container.metadata
         for key in ["build_metadata", "builder", "build_finish", "image"]:
             if key in metadata:
                 del metadata[key]
         return metadata
 
-    @staticmethod
-    def get_download_url(container):
+    def get_download_url(self, container):
 
         secret = container.collection.secret
         download_url = reverse(
@@ -85,12 +82,10 @@ class ContainerSerializer(serializers.HyperlinkedModelSerializer):
     collection = serializers.SerializerMethodField("collection_name")
     metadata = serializers.SerializerMethodField("get_cleaned_metadata")
 
-    @staticmethod
-    def collection_name(container):
+    def collection_name(self, container):
         return container.collection.name
 
-    @staticmethod
-    def get_cleaned_metadata(container):
+    def get_cleaned_metadata(self, container):
         metadata = container.metadata
         for key in ["build_metadata", "builder", "build_finish", "image"]:
             if key in metadata:
@@ -274,8 +269,7 @@ def _container_get(request, container, name=None, tag=None):
 class ContainerSearch(APIView):
     """search for a list of containers depending on a query"""
 
-    @staticmethod
-    def get_object(name, collection, tag):
+    def get_object(self, name, collection, tag):
         from shub.apps.main.query import specific_container_query
 
         return specific_container_query(name=name, collection=collection, tag=tag)
--- a/shub/apps/api/urls/containers.py
+++ b/shub/apps/api/urls/containers.py
@@ -36,17 +36,20 @@ class SingleContainerSerializer(serializers.ModelSerializer):
     image = serializers.SerializerMethodField("get_download_url")
     metadata = serializers.SerializerMethodField("get_cleaned_metadata")
 
-    def collection_name(self, container):
+    @staticmethod
+    def collection_name(container):
         return container.collection.name
 
-    def get_cleaned_metadata(self, container):
+    @staticmethod
+    def get_cleaned_metadata(container):
         metadata = container.metadata
         for key in ["build_metadata", "builder", "build_finish", "image"]:
             if key in metadata:
                 del metadata[key]
         return metadata
 
-    def get_download_url(self, container):
+    @staticmethod
+    def get_download_url(container):
 
         secret = container.collection.secret
         download_url = reverse(
@@ -82,10 +85,12 @@ class ContainerSerializer(serializers.HyperlinkedModelSerializer):
     collection = serializers.SerializerMethodField("collection_name")
     metadata = serializers.SerializerMethodField("get_cleaned_metadata")
 
-    def collection_name(self, container):
+    @staticmethod
+    def collection_name(container):
         return container.collection.name
 
-    def get_cleaned_metadata(self, container):
+    @staticmethod
+    def get_cleaned_metadata(container):
         metadata = container.metadata
         for key in ["build_metadata", "builder", "build_finish", "image"]:
             if key in metadata:
@@ -269,7 +274,8 @@ def _container_get(request, container, name=None, tag=None):
 class ContainerSearch(APIView):
     """search for a list of containers depending on a query"""
 
-    def get_object(self, name, collection, tag):
+    @staticmethod
+    def get_object(name, collection, tag):
         from shub.apps.main.query import specific_container_query
 
         return specific_container_query(name=name, collection=collection, tag=tag)
--- a/shub/apps/api/urls/labels.py
+++ b/shub/apps/api/urls/labels.py
@@ -26,8 +26,7 @@ class LabelSerializer(serializers.ModelSerializer):
 
     containers = serializers.SerializerMethodField("list_containers")
 
-    @staticmethod
-    def list_containers(label):
+    def list_containers(self, label):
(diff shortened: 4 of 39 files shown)


vladdoster:
Mhm.... not quite sure what I am missing here.

Screen Shot 2021-04-03 at 4 04 24 PM

vsoch:
@vladdoster it's usually a mismatch in the version of black - you can check the message in the CI to see what version is installed.

These are great changes! I will try to make some time to review them soon.

Dockerfile
```diff
@@ -41,11 +41,11 @@ RUN apt-get update && apt-get install -y \
     build-essential
 
 # Install Python requirements out of /tmp so not triggered if other contents of /code change
-ADD requirements.txt /tmp/requirements.txt
+COPY requirements.txt /tmp/requirements.txt
```
vsoch, line 44: Wasn't this change already merged?

shub/apps/api/urls/collections.py
```diff
@@ -28,7 +28,8 @@ class CollectionSerializer(serializers.HyperlinkedModelSerializer):
 
     containers = serializers.SerializerMethodField("list_containers")
 
-    def list_containers(self, collection):
+    @staticmethod
+    def list_containers(collection):
```
vsoch, line 32: Did you use some linter to come up with these changes?

I wouldn't change a serializer method field, as it's expecting to have a self (and not be static) see https://www.django-rest-framework.org/api-guide/fields/#serializermethodfield. Did you test these API views with this change?

shub/apps/api/urls/collections.py
```diff
@@ -67,7 +68,8 @@ class CollectionViewSet(viewsets.ReadOnlyModelViewSet):
 class CollectionDetailByName(APIView):
     """Retrieve a collection instance based on it's name"""
 
-    def get_object(self, collection_name):
+    @staticmethod
+    def get_object(collection_name):
```
vsoch, line 72: Same here - I don't think should be changed to static.

shub/apps/api/urls/collections.py
```diff
@@ -95,7 +97,8 @@ class CollectionSearch(APIView):
     a general search to look across all fields for one term
     """
 
-    def get_object(self, query):
+    @staticmethod
```
vsoch, line 100: Same here - I don't think should be changed to static.

shub/apps/api/urls/containers.py
```diff
@@ -36,17 +36,20 @@ class SingleContainerSerializer(serializers.ModelSerializer):
     image = serializers.SerializerMethodField("get_download_url")
     metadata = serializers.SerializerMethodField("get_cleaned_metadata")
 
-    def collection_name(self, container):
+    @staticmethod
```
vsoch, line 39: Same here - I don't think should be changed to static.

shub/apps/api/urls/containers.py
```diff
@@ -36,17 +36,20 @@ class SingleContainerSerializer(serializers.ModelSerializer):
...
+    def collection_name(container):
```
vsoch, line 40: Same here - I don't think should be changed to static.

shub/apps/api/urls/containers.py
```diff
@@ -36,17 +36,20 @@ class SingleContainerSerializer(serializers.ModelSerializer):
...
         return container.collection.name
 
-    def get_cleaned_metadata(self, container):
+    @staticmethod
+    def get_cleaned_metadata(container):
```
vsoch, line 44: Same here - I don't think should be changed to static.

shub/apps/api/urls/containers.py
```diff
@@ -36,17 +36,20 @@ class SingleContainerSerializer(serializers.ModelSerializer):
...
         metadata = container.metadata
         for key in ["build_metadata", "builder", "build_finish", "image"]:
             if key in metadata:
                 del metadata[key]
         return metadata
 
-    def get_download_url(self, container):
+    @staticmethod
```
vsoch, line 51: Same here - I don't think should be changed to static.

shub/apps/api/urls/containers.py
```diff
@@ -82,10 +85,12 @@ class ContainerSerializer(serializers.HyperlinkedModelSerializer):
     collection = serializers.SerializerMethodField("collection_name")
     metadata = serializers.SerializerMethodField("get_cleaned_metadata")
 
-    def collection_name(self, container):
+    @staticmethod
+    def collection_name(container):
```
vsoch, line 89: Same here - I don't think should be changed to static.

shub/apps/api/urls/containers.py
```diff
@@ -82,10 +85,12 @@ class ContainerSerializer(serializers.HyperlinkedModelSerializer):
...
         return container.collection.name
 
-    def get_cleaned_metadata(self, container):
+    @staticmethod
+    def get_cleaned_metadata(container):
```
vsoch, line 93: Same here - I don't think should be changed to static.

shub/apps/api/urls/containers.py
```diff
@@ -269,7 +274,8 @@ def _container_get(request, container, name=None, tag=None):
 class ContainerSearch(APIView):
     """search for a list of containers depending on a query"""
 
-    def get_object(self, name, collection, tag):
+    @staticmethod
+    def get_object(name, collection, tag):
```
vsoch, line 278: Same here - I don't think should be changed to static.

shub/apps/api/urls/labels.py
```diff
@@ -26,7 +26,8 @@ class LabelSerializer(serializers.ModelSerializer):
 
     containers = serializers.SerializerMethodField("list_containers")
 
-    def list_containers(self, label):
+    @staticmethod
+    def list_containers(label):
```
vsoch, line 30: To not be hugely redundant, we should not make any of the rest framework methods staticmethods.  We should follow the example they show in their documentation. To not be redundant I won't comment again, but apply this comment to all following methods.

shub/apps/base/sitemap.py
```diff
@@ -23,7 +23,8 @@ def location(self, obj):
 class ContainerSitemap(BaseSitemap):
     changefreq = "weekly"
 
-    def lastmod(self, obj):
+    @staticmethod
```
vsoch, line 26: Please don't change these to static either. We are following the Django example: https://docs.djangoproject.com/en/3.1/ref/contrib/sitemaps/

shub/apps/base/views/main.py
```diff
@@ -44,7 +44,8 @@ class VersionView(APIView):
 
     renderer_classes = (JSONRenderer,)
 
-    def get(self, request):
+    @staticmethod
```
vsoch, line 47: Please don't add static methods to this (and all following) APIViews. https://stackoverflow.com/questions/49992869/use-staticmethod-in-django-rest-framework-endpoint-method

shub/apps/library/views/helpers.py
```diff
@@ -189,9 +188,9 @@ def generate_collection_tags(collection):
     latest. Technically, collections should be namespaced
     consistently, and we assume this.
     """
-    unique_tags = set(
-        [c.tag for c in collection.containers.all() if not re.search(tag_regex, c.tag)]
-    )
+    unique_tags = {
```
vsoch, line 191: This needs to be a set to ensure we have uniqueness.

shub/apps/main/views/download.py
```diff
@@ -123,15 +123,15 @@ def _download_container(container, request):
         return HttpResponseForbidden()
 
     # A remove build will store a metadata image url
-    elif "image" in container.metadata:
+    if "image" in container.metadata:
```
vsoch, line 126: This needs to be elif, as there could be containers (that had a registry storage switch) that have both "image" in the container.metadata and container.image, and we most certainly don't want both to run.

shub/apps/main/views/download.py
```diff
@@ -141,6 +141,4 @@ def _download_container(container, request):
 
         # There is no container
         raise Http404
```
vsoch, line 143: We probably don't need two here :) Did you run some automated linter to make these changes?

shub/logger.py
```diff
@@ -82,7 +83,8 @@ def emitError(self, level):
             return True
         return False
 
-    def emitOutput(self, level):
+    @staticmethod
```
vsoch, line 86: These static methods added to the logger functions are appropriate.

shub/plugins/pgp/models.py
```diff
@@ -48,7 +48,8 @@ def class_prepared(self, sender, **kwargs):
         post_save.connect(self.post_save, sender=self.model)
         post_delete.connect(self.post_delete, sender=self.model)
 
-    def post_save(self, sender, instance, created, **kwargs):
+    @staticmethod
```
vsoch, line 51: I don't think we should change these either, these follow the suggested way to implement these callbacks.

shub/plugins/pgp/models.py
```diff
@@ -180,13 +182,16 @@ class PGPPacketModel(models.Model):
     class Meta:
         abstract = True
 
-    def is_public_key(self):
+    @staticmethod
```
vsoch, line 185: Could you please test that these changes do not break the keystore?

We should probably discuss tests - if you want to more work on singularity registry it would be a very good use of time to get a test framework setup.

shub/plugins/pgp/views.py
```diff
@@ -96,25 +96,22 @@ def lookup(request):
...
+                content_type="text/plain",
+            )
+            return resp
+
+        # html response
+        op = op if op else "index"
+        if op == "get":
+            c = {"key": utils.keys_ascii_armor(keys), "search": search}
+            return render(request, "pgpdb/lookup_get.html", c)
+        if op in ["index", "vindex"]:
```
vsoch, line 110: I would suggest not changing to if here - if we've already confirmed op == get why should we check it again?

vsoch (commented):
These are great changes @vladdoster ! Wow I didn't know the difference between instantiating a named type (e.g., dict()) vs. the code equivalent ({}) was so large, at least relatively speaking when it scales. You did a really good job removing extra if / else (I don't think I do this commonly anymore but it's still there) and I've added some comments about where to use staticmethod - I think for known Django patterns we should generally follow their instruction, but otherwise (for our own custom models) it should be ok.

Did you use some kind of tool to point to these changes? I'd be interested to know so i can try it out! Maybe pycharm?

vsoch:
Oh and don't worry too much about black - when we're finished up review I'd be happy to pull, run the matching version, and update the PR.

shub/apps/library/views/helpers.py
vladdoster: Yes, `{ }` is the set constructor

vsoch: Totally did not realize that, my mind went to dict but you’re right, there aren’t key value pairs. Thanks for teaching me something new, this is ok then! :)

shub/apps/main/views/download.py
vladdoster: But doesn't the first `if` return? So if goes down that branch it would never reach the next `if` branch?

vsoch: Yes you’re right! Sorry I shouldn’t review when I’m tired 😴

shub/plugins/pgp/models.py
vladdoster: Agreed.

vsoch:
Taking a look again! @vladdoster for future reviews could you wait until the end to rebase? it makes it hard to see what you changed (and I'll need to review the entire thing again instead of just changes).

shub/plugins/google_build/utils.py
```diff
@@ -79,7 +79,7 @@ def paginate(url, headers, min_count=30, start_page=1, params=None, limit=None):
     start_page: the starting page
     """
     if params is None:
```
vsoch, line 81: A trick I do now for a pattern like this is:

```python
params = params or {}
```

shub/plugins/google_build/utils.py
```diff
@@ -283,7 +283,7 @@ def generate_signed_url(storage_path, expiration=None, headers=None, http_method
     credential = "{}/{}".format(client_email, credential_scope)
 
     if headers is None:
-        headers = dict()
+        headers = {}
```
vsoch, line 286: Could work here too!

```python
headers = headers or {}
```

vsoch:
okay I still see a lot of the static methods applied - let me know when this is ready again for review!

PR #359, opened 2021-04-04
(maint) apply quoting to shell scripts

vladdoster:
- add quotes to variables to prevent globbing & word splitting
- remove braces where it's not needed or adds verbosity
- add `x` executable bit to `generate_cert.sh` & `prepare_instance.sh`
- remove HTTPS cert generation scripts using
  [tiny-acme](https://github.com/diafygi/acme-tiny) in favor of
  [CertBot](https://github.com/certbot/certbot)


scripts/generate_certs.sh
```diff
@@ -13,10 +13,10 @@ COUNTY=${4:-San Mateo County}
 sudo mkdir /opt/acme_tiny
```
vsoch, line 13: I don't use tiny acme anymore (I tend to use certbot) - is this example script worth keeping?

vladdoster: I don't think it is worth having. Glancing at the [tiny-acme repository](https://github.com/diafygi/acme-tiny) gives off the impression that it's in maintenance mode.

Screen Shot 2021-04-04 at 2 47 08 AM

Certbot on the other hand is still actively being developed. 

In addition, Certbot is backed by the [EFF](https://www.eff.org/pages/eppcode) whereas tiny-acme is owned by an individual.

vsoch: ok! I'm good removing `generate_certs.sh` and `generate_certs-tiny-acme.sh` (the two that don't use certbot). It looks like we have instructions for cerbot in the rendered logs :point_right: https://singularityhub.github.io/sregistry/docs/install/https

vladdoster: Removed

vladdoster:
Not sure if it is clear in my commit history, but I force-pushed to fix a typo in my commit message. No code changes were made [7093cff](https://github.com/singularityhub/sregistry/commit/7093cff7e56d7137d89ed3b2a4b26a33e575091e).

vsoch (approved):

Issue #377, opened 2021-07-22
Database connection isn't set to UTC


craigwindell:
**Describe the bug**
Commit #f41371a

Accessing the home page of the server results in an error

```
uwsgi_1      | Internal Server Error: /
uwsgi_1      | Traceback (most recent call last):
uwsgi_1      |   File "/usr/local/lib/python3.6/site-packages/django/contrib/sessions/backends/base.py", line 189, in _get_session
uwsgi_1      |     return self._session_cache
uwsgi_1      | AttributeError: 'SessionStore' object has no attribute '_session_cache'
uwsgi_1      |
uwsgi_1      | During handling of the above exception, another exception occurred:
uwsgi_1      |
uwsgi_1      | Traceback (most recent call last):
nginx_1      | 131.181.240.91 - - [22/Jul/2021:23:24:04 +0000] "GET / HTTP/1.1" 500 850 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.164 Safari/537.36 Edg/91.0.864.71" "-"
uwsgi_1      |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
uwsgi_1      |     response = get_response(request)
uwsgi_1      |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
uwsgi_1      |     response = self.process_exception_by_middleware(e, request)
uwsgi_1      |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
uwsgi_1      |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
uwsgi_1      |   File "/usr/local/lib/python3.6/site-packages/ratelimit/decorators.py", line 30, in _wrapped
uwsgi_1      |     return fn(*args, **kw)
uwsgi_1      |   File "./shub/apps/base/views/main.py", line 24, in index_view
uwsgi_1      |     return render(request, "main/index.html")
uwsgi_1      |   File "/usr/local/lib/python3.6/site-packages/django/shortcuts.py", line 36, in render
uwsgi_1      |     content = loader.render_to_string(template_name, context, request, using=using)
uwsgi_1      |   File "/usr/local/lib/python3.6/site-packages/django/template/loader.py", line 62, in render_to_string
uwsgi_1      |     return template.render(context, request)
[...]

PR #378, opened 2021-07-22
Pin psycopg2-binary to fix UTC error

craigwindell:
The fix for #377 we found was to roll back and pin the psycopg2-binary package to the previous version.

The longer term fix is to update the code to work with later versions to solve the error:
AssertionError: database connection isn't set to UTC


vsoch:
You are so quick! Since this is more than a tiny fix, would you mind updating:

- VERSION to indicate one version bump higher (the last number is usually incremeted)
- CHANGELOG.md to reflect/note this change?

Thank you! And you are super speedy - I was responding to your issue and then realized you had already figured it out :)

craigwindell:
I have updated the files you requested.
-Craig

vsoch (approved):

PR #386, opened 2021-11-17
Main

kkaftan:


kkaftan:
sorry was mistake, meant to push in my own forked repo

vsoch:
Any reason you can't contribute the recipe here?

vsoch:
Or we could make a modular repo for the deployment alongside it.

vsoch:
And if you don't have the bandwidth, I can offer to help!

kkaftan:
There is no particular reason not to share. @hashkeks and I were testing the deployment possibilities with k8s and how it works. So what I tried to push is an unrefined version of our project. We would like to improve it, test it further and provide documentation before make publicly available. As it just a side project it may take some time, but we would be glad to come back to you, when we think its ready. :+1:

vsoch:
Sounds good!

Issue #317, opened 2020-07-10
Kubernetes Helm Testing


odatskov:
Dear Developers,

I am currently testing running a Singularity registry on Kubernetes resources. I have ported the docker compose you have into a [basic Helm chart](https://github.com/odatskov/helm-sregistry). Three questions I have:

- The obvious one: is there already a chart or are you working on one (so as not to duplicate effort)?
-  Based on [storage docs](https://singularityhub.github.io/sregistry/docs/install/server#storage): for recent container versions is /var/www/images still needed to be a persistent storage place?  
- worker redis dependency: hardcoded to be a local service or can also be configured with external?

vsoch:
It’s so great that you are doing this! There isn’t a chart and it would be great to have one. The images folder is kept for backwards compatibility, and I believe might still be used if you upload a container directly in the interface (do you have it running and could test?) And I also don’t see why the redis worked could be configured to be external.

Really looking forward to checking this out!

odatskov:
Most excellent! Thank you for such a quick response. Indeed I am running on local K8s infra and will definitely check the interactive upload.

odatskov:
Tested with persistent storage and GitHub auth. Singularity shub pull seems fine, however the push is giving a bit of trouble: singularity doesn't like the tokenfile (signed token?) and with sregistry push I am getting an exception of "certificate verify failed" (even with SREGISTRY_REGISTRY_NOHTTPS=true). 

In regards to previous comments:
- Interactive webui push: indeed the sif image can be found in /var/www/images/<collection>
- Redis custom url I found can be set through REDIS_URL environment.
- settings/config.py [duplicate](https://github.com/singularityhub/sregistry/blob/master/shub/settings/config.py#L72) of DISABLE_BUILDING or expected?

vsoch:
Ah, I bet I know the issue! For the library endpoint you actually can't use http - so the way that I test locally is by compiling singularity with remote.go set to http. Just change [this line](https://github.com/hpcng/singularity/blob/master/internal/pkg/remote/remote.go#L237) from https to http, and then build. The instructions are in the docs [here](https://singularityhub.github.io/sregistry/docs/client#singularity-push). The disable_building repeat is an error! I have a PR open now and I'll remove the extra one.

vsoch:
Any updates on working on a chart?

odatskov:
I am indeed testing the chart on our infra. Will have results in a bit.

ifelsefi:
[...]

PR #408, opened 2022-10-07
Enable streamlined upgrades with local.py

cpeel:
Improve the k8s deployment experience by making override updates to `settings/local.py` rather than changing the entire `settings/config.py`, like was done in the [helm templates](https://github.com/odatskov/helm-sregistry/blob/master/sregistry/templates/sregistry-cm.yaml#L19) referenced in https://github.com/singularityhub/sregistry/issues/317. A single container can now be built for use in dev/stg/prd environments and use an updated helm manifest to override values that differ in each env. This also improves the dev & upgrade experience by having local settings not pollute code under source control.

Pros/cons of this approach:
* Pro:
  * Backwards compatible (users can continue to make changes in `config.py` if they want).
  * Optional -- if there is no `local.py` everything continues to work.
  * Re-uses established pattern of using a python file that does not exist in the repo (eg `secrets.py`).
* Cons:
  * Changes the suggested location of where configurations are set (`local.py` vs `config.py`) requiring several doc updates.
  * Users must copy/paste/change the settings from `config.py` into `local.py`.

Other approaches considered:
* We could move to a config file or env-based approach. The challenge would be lots of changes to `config.py` to optionally read every variable from the file or env before falling back to a default/initial value. This approach is also challenging for representing non-string / non-numeric configuration settings (booleans, lists).
* All of this is technically not necessary as users can just write config values into `secret.py` and have them picked up and override `config.py`. This hacky approach convolutes the separation of secrets vs configuration data.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,8 +20,8 @@ make_video.sh
 celerybeat-schedule.db
 *.mp4
 OLD
-secrets.py
-local_settings.py
+shub/settings/secrets.py
+shub/settings/local.py
 cronjob
 
 # Static #
--- a/docs/_docs/install/https.md
+++ b/docs/_docs/install/https.md
@@ -83,7 +83,7 @@ The certificate is at "./singularity-registry.org+5.pem" and the key at "./singu
 It will expire on 29 August 2023 🗓
 ```
 
-Then I moved them into the registry root, and updated my shub/settings/config.py to use
+Then I moved them into the registry root, and updated my `shub/settings/local.py` to use
 https on localhost.
 
 ```python
--- a/docs/_docs/install/server.md
+++ b/docs/_docs/install/server.md
@@ -45,8 +45,8 @@ If you don't care about user experience during updates and server downtime, you 
 ## Custom Domain
 
 In the [config settings file](https://github.com/singularityhub/sregistry/blob/master/shub/settings/config.py#L30)
-you'll find a section for domain names, and other metadata about your registry. You will need to update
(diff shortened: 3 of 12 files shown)


vsoch:
> We could move to a config file or env-based approach. The challenge would be lots of changes to config.py to optionally read every variable from the file or env before falling back to a default/initial value. This approach is also challenging for representing non-string / non-numeric configuration settings (booleans, lists).

I think I actually really like this approach? I made the original sregistry a long time ago, but since then I've made (by default) most of my Django apps read everything in from either a config file or the environment. E.g., here is an example that allows loading in from a config.yaml or the environment: 

https://github.com/spack/spack-monitor/blob/89acf94dc664b598d9f73292ae4d61fdccf3ac5b/spackmon/settings.py#L44-L55

And then the defaults could be set in multiple ways, e.g., here is an example (for a Django plugin) that has a defaults structure and then can override it (in our case it would be from a settings file or environment):

https://github.com/vsoch/django-river-ml/blob/67bea67770809e75323a68cac660f528e673f5d2/django_river_ml/settings.py#L31-L102

What do you think?

vsoch:
hey @cpeel ! Here is an idea to throw into the mix for envars: https://github.com/singularityhub/sregistry/pull/409. Its a bit different than what I've done before (I wanted more control over types) but I think it could be a good start? If you want to test out the branch (I won't get to before the work day starts) we can iteratively work on it (if you think it is a good direction).

cpeel:
Closing in lieu of https://github.com/singularityhub/sregistry/pull/409

PR #409, opened 2022-10-07
Allowing setting config vars from environment

vsoch:
This is an idea to address the same issue brought up in #408 - we want to be able to easily set variables in settings from the environment. This currently just includes values in config.py (and currently can be extended). I decided to explicitly list the names and types to have better control (e.g., instead of iterating over locals().


shub/settings/config.py
```diff
@@ -96,11 +115,12 @@
 
 DATABASES = {
     "default": {
-        "ENGINE": "django.db.backends.postgresql_psycopg2",
-        "NAME": "postgres",
-        "USER": "postgres",
-        "HOST": "db",
-        "PORT": "5432",
+        "ENGINE": os.environ.get("SREGISTRY_DATABASE_ENGINE")
```
cpeel, line 118: Consider leveraging the optional default arg of `.get()`:
```python
os.environ.get("SREGISTRY_DATABASE_ENGINE", "django.db.backends.postgresql_psycopg2")
```

shub/settings/config.py
```diff
@@ -159,3 +179,89 @@
...
+
+# Any plugins enabled from the environment?
+PLUGINS_ENABLED += get_sregistry_envar_list("SREGISTRY_PLUGINS_ENABLED")
+
+## Environment
+
+# Custom additions are handled before this
+# Limit defaults that user can set
+BOOLEAN_DEFAULTS = [
+    "ENABLE_GOOGLE_AUTH",
```
cpeel, line 191: Perhaps alpha-sorting these lists?

shub/settings/config.py
```diff
@@ -159,3 +179,89 @@
...
+    "ENABLE_TWITTER_AUTH",
+    "ENABLE_GITHUB_AUTH",
+    "ENABLE_GITLAB_AUTH",
+    "ENABLE_BITBUCKET_AUTH",
+    "ENABLE_GITHUB_ENTERPRISE_AUTH",
+    "USER_COLLECTIONS",
+    "PRIVATE_ONLY",
+    "DEFAULT_PRIVATE",
+    "DISABLE_BUILDING",
+    "SREGISTRY_GOOGLE_BUILD_SINGULARITY_VERSION",
```
cpeel, line 201: A bit unfortunate that this will have 2x `SREGISTRY_` but I can't think of an easy way around it without renaming and that would break backwards compatibility.

cpeel: Also this variable is a string and not a boolean.

shub/settings/config.py
```diff
@@ -159,3 +179,89 @@
...
+    "CONTAINER_SIGNED_URL_EXPIRE_SECONDS",
+    "MINIO_SIGNED_URL_EXPIRE_MINUTES",
+]
+
+# Set envars by type
+for key in BOOLEAN_DEFAULTS:
+    value = get_sregistry_envar(key)
+
+    # Setting a boolean
+    if value in ["true", "True"]:
```
cpeel, line 244: For better robustness, consider:
```python
if value.lower() in ["true", "1"]:
```
That way it works for true, True, TRUE, 1, etc

shub/settings/config.py
```diff
@@ -159,3 +179,89 @@
...
+        locals()[key] = True
+        print("Setting %s from the environment to True" % key)
+
+    elif value in ["False", "false"]:
```
cpeel, line 248: Ditto here:
```python
if value.lower() in ["false", "0"]:
```

shub/settings/config.py
```diff
@@ -159,3 +179,89 @@
...
+        print("Setting %s from the environment to False" % key)
+
+# Integer values
+for key in INTEGER_DEFAULTS:
+    value = get_sregistry_envar(key)
+    if value:
+        try:
+            locals()[key] = int(value)
+        except:
+            print("There was an issue setting %s as an integer." % key)
```
cpeel, line 259: It would be better to either let this fail or re-raise an exception so that the container fails rather than continue with an unexpectedly invalid value.

shub/settings/config.py
```diff
@@ -159,3 +179,89 @@
...
+        locals()[key] = False
+        print("Setting %s from the environment to False" % key)
+
+# Integer values
+for key in INTEGER_DEFAULTS:
+    value = get_sregistry_envar(key)
+    if value:
+        try:
+            locals()[key] = int(value)
+        except:
```
cpeel, line 258: `except ValueError` would be better here. If `int(value)` is returning something besides that it's way better to just barf all over the place.

shub/settings/config.py
```diff
@@ -159,3 +179,89 @@
...
+]
+
+# Set envars by type
+for key in BOOLEAN_DEFAULTS:
+    value = get_sregistry_envar(key)
+
+    # Setting a boolean
+    if value in ["true", "True"]:
+        locals()[key] = True
+        print("Setting %s from the environment to True" % key)
```
cpeel, line 246: I'm on the fence about printing when we set the value. This is really useful when debugging ("arg, I'm setting the stupid thing why isn't it picking it up!") but could also be a bit of noise. I think I'd err on outputting that it's being picked up but probably doing it via `logger.info()`. I do think it would be good if we did it for all variable types and only after we set the variable rather than before, for consistency (see the strings section).

cpeel (commented):
Overall I like this approach! The challenge with doing it here at the end of `config.py` rather than `__init__.py` is that we can't override other variables set in other `settings/*` files, like those in `tasks.py` (which we would want to do in our k8s scenario as we move redis elsewhere).

I also fear that we now need to keep the lists up-to-date when a new variable is added. We could add some CI that checks to see if any variable is in the file that isn't in one of the other lists.

vsoch:
@cpeel totally agree - let me rework this a bit (probably after work) to see if I can make it a neater solution. I don't like the redundancy either!

cpeel:
I'm not sure refactoring it into a single `settings.py` is required. The way you've done the importing via `__init__.py` is nice and it lets them all be "local" to `settings.py` but logically grouped. If you moved the overrides into `__init__.py` after the imports rather than in `config.py` that would ensure that the overrides apply to any file.

One way to decrease the redundancy in this approach might be to iterate over all env settings and pull out all of them with `SREGISTRY_` prefixes. If there is a matching local variable, set it. If it is prefixed by 'ENABLE` or `DISABLE` treat it as a boolean. Otherwise try to cast it as an int, otherwise treat it as a string. This has the potential to be more error prone -- and I would need to review all of the existing settings to see if this would work -- but it's extensible and removes the redundancy. We'd still need to special-case the lists in the code, but those are actually pretty few. An idea anyway.

Another possible approach would be to have the settings in a JSON file that is specified via env variable (`SREGISTRY_CONFIG_FILE`). The JSON would allow the user to be explicit about the type and take care of things like lists, dicts, etc. So the override code would open and iterate over the JSON and set local settings to their JSON values. In k8s we'd just create the JSON file in the manifest and pass in the env variable to pick it up. The more I type the more I like that idea.

vsoch:
> Another possible approach would be to have the settings in a JSON file that is specified via env variable (SREGISTRY_CONFIG_FILE). The JSON would allow the user to be explicit about the type and take care of things like lists, dicts, etc. So the override code would open and iterate over the JSON and set local settings to their JSON values. In k8s we'd just create the JSON file in the manifest and pass in the env variable to pick it up. The more I type the more I like that idea.

I also like the idea of having this as an option, although not required. And I'm also leaning towards making everything very flexible - e.g., "provide the value directly in the file, via the config file, or environment - up to you!" 

Is JSON preferred to YAML? I think (for configs like this) my preference would be YAML, but maybe my config-ometer sense is off. :laughing:

vsoch:
okay will have a PR for you this evening!

cpeel:
> Is JSON preferred to YAML? I think (for configs like this) my preference would be YAML, but maybe my config-ometer sense is off.

YAML is probably better (says my professional half -- my personal half thinks YAML is rubbish 😁 ).

vsoch:
>  YAML is probably better (says my professional half -- my personal half thinks YAML is rubbish grin ).

Haha, I totally feel that!

So I just pushed a refactor that has a single settings.py (to give more confidence we find everything in one place) and it should work to define variables from:

- defaults in that file (e.g., a local or custom build)
- a settings.yaml in the root OR a custom path (e.g., still local or cloud / Kubernetes)
- from the environment (could also work in either case!)
- still a secrets.py alongside the settings.py (I realize I enabled `SECRET_KEY` and we might need to move this import higher up to ensure it doesn't get overridden!)
 
https://github.com/singularityhub/sregistry/pull/409/commits/9b25c8cd26fa62c2f7af2d2866167b656d1bfd1b

I decided to still have tight control over variables and placed them into groups based on type, as oppose to a strategy that looks for any `SREGISTRY_` prefix in the environment and then tries to guess. It's an improvement on my first shot because we remove the redundancy. If we want to expose more customization (e.g., in the various dicts in settings.py) that should be fairly easy to do. 

The one change I'm still thinking about doing is not having any "special prefix" for secrets as I do now (e.g., `SREGISTRY_SECRET_`. I was thinking that might be useful for the developer - so they can see based on the name that something should be treated more carefully. I'm thinking maybe it makes sense to try out this branch for your use case, and we can tweak / change as you see issues arise?

shub/settings.py
```diff
@@ -0,0 +1,537 @@
...
+    "DOMAIN_NAME_HTTP": "http://127.0.0.1",
+    # Get additional admins from the environment
+    "HELP_CONTACT_EMAIL": "vsoch@users.noreply.github.com",
+    "HELP_INSTITUTION_SITE": "https://srcc.stanford.edu",
+    "REGISTRY_NAME": "Tacosaurus Computing Center",
+    "REGISTRY_URI": "taco",
+    # "UA-XXXXXXXXX"
+    "GOOGLE_ANALYTICS": None,
+    "GOOGLE_BUILD_SINGULARITY_VERSION": "v3.3.0-slim",
+    "SREGISTRY_DATABASE_ENGINE": "django.db.backends.postgresql_psycopg2",
```
cpeel, line 85: We should consider not prefixing these with `SREGISTRY_` here so that they don't have to be double-prefixed in the environment variables.

shub/settings.py
```diff
@@ -0,0 +1,537 @@
...
+    # Minio Storage
+    "MINIO_SERVER": "minio:9000",  # Internal to sregistry
+    "MINIO_EXTERNAL_SERVER": "127.0.0.1:9000",  # minio server for Singularity to interact with
+    "MINIO_BUCKET": "sregistry",
+    "MINIO_REGION": "us-east-1",
+    # The rate limit for each view, django-ratelimit, "50 per day per ipaddress)
+    "VIEW_RATE_LIMIT": "50/1d",
+    "DJANGO_LOG_LEVEL": "WARNING",
+    # An image url or one of the following: 'mm', 'identicon', 'monsterid', 'wavatar', 'retro'. Defaults to 'mm'
+    "GRAVATAR_DEFAULT_IMAGE": "retro",
```
cpeel, line 99: The comment says the default is `mm` but we are setting `retro` here.

shub/settings.py
```diff
@@ -0,0 +1,537 @@
...
+
+# Environment helpers
+
+
+def get_sregistry_envar(key):
+    envar_key = "SREGISTRY_%s" % key
+    return os.environ.get(envar_key)
+
+
+def get_sregistry_envar_list(key):
```
cpeel, line 129: The only 2 places where this is used already prefix `key` with `SREGISTRY_`. Consider doing the prefixing within this function like you do for `get_sregistry_envar()` so they are parallel and the caller doesn't need to prefix it.

shub/settings.py
```diff
@@ -0,0 +1,537 @@
...
+    if envar_list:
+        envar_list = [x.strip() for x in envar_list.split(",") if x.strip()]
+    return envar_list
+
+
+# Try reading in from secrets first
+try:
+    from .secrets import *
+
+# Fallback to dummy secrets (usually for testing)
```
cpeel, line 140: This would be surprising for me as a user and I would discourage us from importing an example file. It could mask a problem during development that would raise itself during production deployment.

shub/settings.py
```diff
@@ -0,0 +1,537 @@
...
+        "NAME": cfg.SREGISTRY_DATABASE_NAME,
+        "USER": cfg.SREGISTRY_DATABASE_USER,
+        "HOST": cfg.SREGISTRY_DATABASE_HOST,
+        "PORT": cfg.SREGISTRY_DATABASE_PORT,
+    }
+}
+
+# STORAGE
+
+MINIO_ROOT_USER = os.environ.get("MINIO_ROOT_USER")
```
cpeel, line 343: While we're here, why not allow this to be set via the config as well? To keep backwards compatibility we could do:
```
MINIO_ROOT_USER = os.environ.get("MINIO_ROOT_USER", cfg.MINIO_ROOT_USER)
```
and make the default for `MINIO_ROOT_USER` in the `STRING_DEFAULTS` to be `None`.

shub/settings.py
```diff
@@ -0,0 +1,537 @@
...
+# https://docs.djangoproject.com/en/1.9/howto/static-files/
+
+MEDIA_ROOT = "/var/www/images"
+MEDIA_URL = "/images/"
+STATIC_ROOT = "/var/www/static"
+STATIC_URL = "/static/"
+UPLOAD_PATH = "%s/_upload" % MEDIA_ROOT
+
+
+RQ_QUEUES = {"default": {"URL": os.getenv("REDIS_URL", "redis://redis/0")}}
```
cpeel, line 476: We need to be able to set the redis config via YAML/env as well. Although the way you're reading the YAML and setting the locals at the end of this file it looks like that implicitly works for YAML already.

requirements.txt
```diff
@@ -42,3 +42,4 @@ social-auth-core[saml]
 sregistry[all-basic]>=0.2.19
 uwsgi
 minio==5.0.8
+pyyaml
```
cpeel, line 45: Consider keeping these alpha-sorted (although I guess `minio` isn't alpha sorted here either).

shub/settings.py
```diff
@@ -0,0 +1,537 @@
...
+
+# Finally, ensure all variables in cfg are set in locals
+for key, value in cfg:
+    locals()[key] = value
+
+# Add SREGISTRY_SECRET_ from the environment
+for key, value in os.environ.items():
+    if key.startswith("SREGISTRY_SECRET_"):
+        secret_key = key.replace("SREGISTRY_SECRET_", "")
+        if not value:
```
cpeel, line 531: You want `if value is not None:` here, otherwise if `value` is "falsey" like `0` it will not be set.

cpeel (commented):
> The one change I'm still thinking about doing is not having any "special prefix" for secrets as I do now (e.g., `SREGISTRY_SECRET_`).

Yes, I was thinking that too after reviewing the code. I don't think prefixing `SECRET` makes the developer any more (or less) aware of the importance of that and think it would be simpler overall if we didn't do the special prefixing. And as a counter-example we're not prefixing that to the postgres DB password but that's arguably also important.

----

While I haven't yet tested this, the concept is solid and I think this will work for what we need to do.

One thing I am wary of is that this will break backwards compatibility for users who upgrade and in the past may have been replacing `CONFIG.py` with their settings. This also renames some configuration variables. I would classify this is an "API-breaking change" and would rev this as a major point release, rather than a fix update, per [semver](https://semver.org/)

shub/settings.py
vsoch: Agree.

shub/settings.py
vsoch: I'll clarify this - I think if it's unset it defaults to mm, and we override with retro.

shub/settings.py
vsoch: The only thing defined in the dummy secrets (after I tweak it) would be the `SECRET_KEY` - and this is mostly done to ensure that if someone clones the repository, does nothing, and brings it up with docker-compose there will be a secret key.  I think what I'm going to do is move the entirety of secrets to be alongside the main settings.py (since they are optional and can come from the environment) and then keep the dummy_secrets.py, but print a huge warning about the secret key not being modified. What do you think? Is there another or a better way to go about this?

shub/settings.py
vsoch: We could definitely do that, although for the actual Minio server it is expecting `MINIO_ROOT_USER` explicitly. I think the use case that you have in mind is having deployed your Minio separately (and so you could use `SREGISTRY_MINIO_ROOT_USER` without an issue? I'm definitely happy to add that!

vsoch: I also realize that I need to do a check if a value is set before setting to locals, e.g., for the case that the cfg MINIO_ROOT_USER (from `SREGISTRY_MINIO_ROOT_USER) is None, we don't want it to override `MINIO_ROOT_USER`

shub/settings.py
cpeel: > and this is mostly done to ensure that if someone clones the repository, does nothing, and brings it up with docker-compose there will be a secret key

I would assert that it's better to hard fail with an exception if that is required and not set rather than using a testing value. That makes it both more secure and doesn't silently fail (or even worse: be used insecurely).

shub/settings.py
cpeel: Correct - I need to decouple the two. Eventually I need this to work against a genuine S3 bucket rather than a minio server, but I haven't yet gotten to testing that / making changes to support it.

vsoch: Oh nice! That would be amazing to have (and based on the shared protocol hopefully not too hard to support). I can definitely help when the time comes.

shub/settings.py
vsoch: How about:

```
# If we don't have a secret key, fire a warning
if "SECRET_KEY" not in locals():
    print("****** WARNING - using dummy secret key - you MUST change for production! ******")
    SECRET_KEY = "xxxxxxxxxxxxxxxxxx"
```
I understand the concern, but people's first impression of a piece of a software are hugely influenced by these interactions, and I think we can find a balance by putting enough verbosity alongside this (and ensuring there are ample docs, etc. that tell the person to set it).

cpeel: I get where you're coming from, but that makes it insecure by default and console warnings are way too easy to overlook.

Is it better to work out-of-the-box but risk a security vulnerability in production or start with a solid security posture and loudly fail with an error on how to fix it out-of-the-box?

And as a final data point, the current version doesn't work out-of-the-box without this set, so we don't "lose" anything by maintaining that approach.

vsoch: okay that's fair! I'll adjust to:

```python
# If we don't have a secret key, no go
if "SECRET_KEY" not in locals():
    sys.exit("SECRET_KEY is required but not set. Set SREGISTRY_SECRET_KEY.")
```

shub/settings/config.py
vsoch: I've actually had times when it's set (but empty) so what I typically do to get around that is:

```python
os.environ.get('key') OR "default"
```

shub/settings/config.py
vsoch: ah missed this! The current ordering (in settings.py) is my best effort to have them grouped logically by function (within a particular variable type, and over alpha sorting). Let me know if you think this should be done differently.

shub/settings/config.py
vsoch: oh missed these - good point we should allow booleans to also be set as 0/1.

shub/settings/config.py
cpeel: Your current version where they are logically grouped (with comments!) is far better and we should keep what you have now.

vsoch:
okay (barring no issues with building) another round of changes! https://github.com/singularityhub/sregistry/pull/409/commits/854b7ab9be83c44f4ded920d7509cc43ca155b1b

The only detail we have to be careful about now is that values that are expected to be in settings (e.g., `GOOGLE_ANALYTICS` that can still be set to undefined are provided as empty string `""` and not None - Nones are assumed to be unset and won't be added to the settings locals, whereas an empty string will.

docs/_docs/install/settings.md
```diff
@@ -6,73 +6,119 @@ toc: false
...
+For any setting in the sections below, you can set them in the environment (with `SREGISTRY_`)
+or via the settings.yaml file or the `secrets.py`. If you choose to use a file (and are working locally) you'll want to start by copying
+either of the templates:
 
 ```bash
-touch shub/settings/secrets.py
+cp shub/dummy_secrets.py shub/secrets.py
+cp shub/dummy-settings.yaml settings.yaml
 ```
+And make sure neither of these files are added to any kind of version control!
```
cpeel, line 69: Oh, good point. We should update `.gitignore` to now include `settings.yaml` to ignore.

shub/settings.py
```diff
@@ -0,0 +1,662 @@
...
+    value = get_sregistry_envar(key)
+
+    if value is None:
+        continue
+
+    # Setting a boolean
+    if value.lower() in ["true", "1"]:
+        BOOLEAN_DEFAULTS[key] = True
+
+    elif value.lower() in ["false", "1"]:
```
cpeel, line 237: 1 should be 0 here.

shub/settings.py
```diff
@@ -0,0 +1,662 @@
...
+        BOOLEAN_DEFAULTS[key] = False
```
cpeel, line 238: Should we add an:
```python
else:
    sys.exit("There was an issue setting %s as a boolean." % key)
```
to mirror the integer pattern?

docs/_docs/install/settings.md
```diff
@@ -6,73 +6,119 @@ toc: false
...
+As of version `2.0.0` we have one core [settings.py](https://github.com/singularityhub/sregistry/blob/master/shub/settings.py) file
+that can be populated in several ways:
+
+ - Directly in the file (e.g., if you build your own base container)
+ - From the environment (described below with a `SREGISTRY_` prefix
+ - From a config file, `settings.yaml` in the root directory of your server OR custom set at `SREGISTRY_SETTINGS_FILE`
+ 
+Order of preference or variables honored is:
+
+ 1. secrets.py
```
cpeel, line 18: The way it is coded now, the order is:
1. secrets.py
2. settings.yaml
3. the environment
4. defaults directly in settings.py

Which I think is the right order, because we:
1. set the defaults
2. override any values from the env
3. override any values from settings.yaml
4. only override values already set in the file from `secrets.yaml`.

shub/settings.py
```diff
@@ -0,0 +1,662 @@
...
+
+def get_sregistry_envar_list(key):
+    key = "SREGISTRY_%s" % key
+    envar_list = os.environ.get(key) or []
+    if envar_list:
+        envar_list = [x.strip() for x in envar_list.split(",") if x.strip()]
+    return envar_list
+
+
+# Try reading in from secrets first (no issue if not found)
```
cpeel, line 220: I would move this to the end of the file so that the load order maps to the conceptual "define and then override" that we're doing:
1. defaults
2. environment
3. settings.yaml
4. secrets.py

The secrets currently take precedence because of this code on line 664:
```python
    # Don't set if the value is empty, or it's been set previously
    if value is None or key in locals() and locals()[key] is not None:
```

I think the order you have is correct and moving the secrets imports to line 648 makes it very clear that the file trumps everything else.

cpeel (commented):
This is looking 💯 . Some small PR feedback. Still need to do some testing on it.

shub/settings.py
vsoch: haha oops! I'll update these in a few minutes - finishing up something else!

docs/_docs/install/settings.md
vsoch: Environment (starting on line 226) actually comes before settings.yaml (275) , however we could switch this around so the settings.yaml is first. I was thinking that the secrets.py should take first preference for backwards compatibility, and then environment should always override everything else (e.g., a config file might already exist but the user can easily set the envar to override it) and then the defaults in settings.py are last. What do you think?

shub/settings.py
vsoch: Ah ok, so you agree about environment being honored first? I think my point of confusion is that I'm loading in this order:

1. secrets.py
2. environment
3. settings.yaml
4. defaults

and the implementation does that by using that order, but never setting something that is already defined.

vsoch: Agree - we can put secrets at the bottom, although I'll want to double check / think about if any of the previous settings (e.g., putting a value into a data structure) would prevent that.

vsoch: okay I think this will work - with the assumption that secrets.py will work only for values you could previously put there (e.g., the new values for the various API settings that I'm adding newly will not work) and to over-ride entire datastructures (e.g., the database) you need to define the entire thing (and it's not exposed just via attributes of that).

docs/_docs/install/settings.md
cpeel: Yes, sorry, I read the code wrong. What you have here in `settings.md` matches the behavior in `settings.py`.

It might be better if the env overrides the `settings.yaml` file, but I'm not going to lose any sleep over what you have here.

vsoch: I do agree environment should not be over-ridden - the challenge here is that we don't have a good way to determine if a variable was set in the environment (or is still a default). We can't just check if a DEFAULT[key] is not None, because this won't work for booleans, or generally any variable with a default that isn't None. It's a bit of a catch 22 because in order to set anything from settings.yaml, we need to have all the variables in one lookup, `DEFAULTS`, but we can't set that one variable until we've parsed the different types from the environment! So I think (for now at least) this approach is a reasonable start. If you have a good idea for an implementation though I'd definitely be down to try it!

vsoch:
Note to self: don't look back at old code, because it undoubtedly will look terrible and you'll want to change everything to make it look nicer! :sob: (true story, just now...) :laughing:

cpeel:
> Note to self: don't look back at old code, because it undoubtedly will look terrible and you'll want to change everything to make it look nicer! 😭 (true story, just now...) 😆

lol - I've been doing software development for 22 years and this only gets worse! On the plus side this means that we're becoming better software developers!

> Thanks for the back and forth today @cpeel it was fun!

Same! Appreciate all of your work in this PR! I will give this a test run ASAP. Tomorrow is busy for me so might be Tuesday.

shub/settings.py
```diff
@@ -0,0 +1,669 @@
...
+    pass
+
+if not MINIO_ROOT_USER or not MINIO_ROOT_PASSWORD:
+    print(
+        "Warning: either MINIO_ROOT_USER or MINIO_ROOT_PASSWORD is missing, storage may not work."
+    )
+
+# If we don't have a secret key, no go
+if "SECRET_KEY" not in locals():
+    sys.exit("SECRET_KEY is required but not set. Set SREGISTRY_SECRET_KEY.")
```
cpeel, line 658: Setting `SECRET_KEY` in the YAML file works fantastically, but setting `SREGISTRY_SECRET_KEY` env var doesn't work because we don't define it in the `STRING_DEFAULTS`. It's probably easiest to keep the functionality the way it is now and just update the message to `Set SECRET_KEY in secrets.py or a YAML config file.`

shub/settings.py
```diff
@@ -0,0 +1,669 @@
...
+    # Can leave blank if anonymous binding is sufficient
+    "AUTH_LDAP_BIND_DN": "",
+    "AUTH_LDAP_BIND_PASSWORD": "",
+    "AUTH_LDAP_USER_SEARCH": None,  # "ou=users,dc=example,dc=com"
+    "AUTH_LDAP_GROUP_SEARCH": None,  # "ou=groups,dc=example,dc=com"
+    # Anyone in this group can get a token to manage images, not superuser
+    "AUTH_LDAP_STAFF_GROUP_FLAGS": None,  # "cn=staff,ou=django,ou=groups,dc=example,dc=com",
+    # Anyone in this group is a superuser for the app
+    "AUTH_LDAP_SUPERUSER_GROUP_FLAGS": None,  # "cn=superuser,ou=django,ou=groups,dc=example,dc=com"
+    # "cn=sregistry_admin,ou=groups,dc=example,dc=com"
```
cpeel, line 162: Not sure what this LDAP string is for -- is it a second example for `AUTH_LDAP_SUPERUSER_GROUP_FLAGS` above?

cpeel (approved):
This is working great!

Found two small minor comments in my testing but this looks good to go on my end.

shub/settings.py
vsoch: yes, just a second example! I had two previously and wanted to keep both of them for the developer user. I'm not great with LDAP so I couldn't tell you the difference, I'd have to ping the plugin contributor to get details.

shub/settings.py
vsoch: Any reason to not allow this in the environment? I am thinking we can just add to the `STRING_DEFAULTS` and be more consistent?

cpeel: No reason I can think of. Adding it to `STRING_DEFAULTS` and updating the check for it to be `not None` makes sense.

shub/settings.py
cpeel: I would remove this line then as I don't think it provides any additional information for someone who is integrating with LDAP -- the other one is sufficient. (Says someone who worked on LDAP code for 10 years.)

vsoch:
@cpeel I'll get these final tweaks in tonight - I don't typically commit to personal projects during the work day!

vsoch:
okay final two changes in! https://github.com/singularityhub/sregistry/pull/409/commits/b9335b6d7f29583579daccdbdc798419228daf5b

cpeel (approved):
Ship it!

PR #417, opened 2023-03-27
Remove wrong information

SethosII:
Automated lookup attributed false information.


vsoch:
@SethosII are you able to replace it with the correct information? My worry is that it will see it missing and look up the incorrect info again!

vsoch:
Let’s try it! Tributors will run after merge and we can see if it tries to update again. If so, I might need to tweak the library to not do that.

PR #418, opened 2023-03-27
Add missing package

SethosII:
Package python-jose is required for LDAP authentication.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -31,6 +31,7 @@ Pillow
 psycopg2-binary~=2.8.6
 pygments
 python3-saml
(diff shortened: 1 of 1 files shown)


vsoch:
Should this perhaps be here as another line https://github.com/singularityhub/sregistry/blob/1cce4be2812eecf4ee0be76da9d6e8078b4bdfa1/Dockerfile#L55 if it’s only needed for LDAP?

vsoch:
That works for me then - it's worrisome they don't install one of their dependencies and put that burden on us.

Issue #416, opened 2023-03-24
LDAP login: Python package jose missing, Python syntax error during login


SethosII:
**Describe the bug**

I try to setup LDAP login with Active Directory. There is the `Beep Boop, Error!` page after I enter my credentials and click on login. The logs show this:

```
uwsgi_1      |   File "/usr/local/lib/python3.9/site-packages/social_core/backends/open_id_connect.py", line 5, in <module>
uwsgi_1      |     from jose import jwk, jwt
uwsgi_1      | ModuleNotFoundError: No module named 'jose'
```

Therefore I added `jose` to the `requirements.txt` and rebuilt the image. Now I still get the `Beep Boop, Error!` page after I enter me credentials and click on login, but the logs show this:

```
uwsgi_1      |   File "/usr/local/lib/python3.9/site-packages/social_core/backends/open_id_connect.py", line 5, in <module>
uwsgi_1      |     from jose import jwk, jwt
uwsgi_1      |   File "/usr/local/lib/python3.9/site-packages/jose.py", line 546
uwsgi_1      |     print decrypt(deserialize_compact(jwt), {'k':key},
uwsgi_1      |           ^
uwsgi_1      | SyntaxError: invalid syntax**
```

So this seeams to be a Python error not a configuration error. Also, `jose` seems to be outdated ([last release on pypi 2015](https://pypi.org/project/jose/#history)) so probably it doesn't work with Python 3. Do you have any ideas how to fix this?

**To Reproduce**

Setup `sregsitry` for LDAP login and try to login.

I followed these steps:

```
sudo apt install docker.io docker-compose
git clone https://github.com/singularityhub/sregistry
cd sregistry/
cp shub/dummy-settings.yaml settings.yaml
vi settings.yaml (I modified SECRET_KEY, DOMAIN_NAME, DOMAIN_NAME_HTTP, REGISTRY_NAME, REGISTRY_URI, AUTH_LDAP_SERVER_URI, AUTH_LDAP_BIND_DN, AUTH_LDAP_BIND_PASSWORD, AUTH_LDAP_USER_SEARCH, AUTH_LDAP_GROUP_SEARCH, AUTH_LDAP_SUPERUSER_GROUP_FLAGS
vi Dockerfile (I modified ARG ENABLE_LDAP)
vi shub/settings.py (I modified ldap_auth)
sudo docker build --build-arg ENABLE_LDAP=true -t quay.io/vanessa/sregistry .
sudo docker-compose up
[...]

PR #419, opened 2023-03-27
Small documentation fixes

SethosII:
Fix documentation errors mentioned in https://github.com/singularityhub/sregistry/issues/416#issuecomment-1485078104.


vsoch:
The path to ldap.png is correct. If it's not showing up you could try adding site.baseurl or similar to get the full path, although then it won't render in the markdown.

SethosII:
@vsoch For me it looks like this in Firefox/Chrome/Edge: 
ldap_image_broken

I use this way to include images in my Jekyll sites: `/sregistry/docs/plugins/ldap/ldap.png` but [other images in the repository](https://github.com/singularityhub/sregistry/search?p=4&q=png) use something like this for the paths: `../../assets/img/bitbucket-consumer.png`.

By the way, should the ldap image also be in `assets/img`?

vsoch:
If that works for the other images, moving it there should be a good idea! 👍

SethosII:
@vsoch I also changed it for the images of the globus page.

vsoch:
Did you run locally to confirm they are all visible? That's the last check I'd do - once you confirm that we can merge I think!

vsoch:
@SethosII I usually use jekyll from rb-env (and I haven't installed for a while) so I suspect the install might vary. If it was hard /annoying enough and you want to tweak the docs please feel free to open another PR, but it's not required!

PR #435, opened 2023-06-02
allow setting of gitlab scope from env or settings file

### LINKED ISSUE #431: PLUGINS_ENABLED cannot be set in the settings.yaml file.
the PLUGINS_ENABLED parameter can only be set in the environment, and not the settings.yaml file.  I also suspect the secrets.py file will also be ignored.

https://github.com/singularityhub/sregistry/blob/ee23901a00a23d0d2a65fab5f484c803cfd60e44/shub/settings.py#L441

### LINKED ISSUE #432: config.py is no longer there.
Since #409 removed the settings directory, the code:

https://github.com/singularityhub/sregistry/blob/ee23901a00a23d0d2a65fab5f484c803cfd60e44/run_uwsgi.sh#L14

will never be able to find the file, It may be in the settings.yaml file in the shub directory, or possibly in the environment?

### LINKED ISSUE #434: hard coding SOCIAL_AUTH_GITLAB_SCOPE breaks auth.
In the settings file if gitlab is enabled, the code just forces SOCIAL_AUTH_GITLAB_SCOPE to be api and read_user scopes.

https://github.com/singularityhub/sregistry/blob/ee23901a00a23d0d2a65fab5f484c803cfd60e44/shub/settings.py#L671

This seems to break my auth, as the social auth adds `read_user` onto the end of what is asked for, effectively causing a duplicate in the scopes that are handed to gitlab.  This seems to break gitlab auth with a `The requested scope is invalid, unknown, or malformed.` message

I also don't need the `api` scope just for auth, and there is no way to provide an override setting since the code just sets it as the last thing.

My suggestion is to:
make the default `[]`, and provide a way to set it from the environment or the settings.yaml file.

Discuss...

karcaw:
This changes the default of SOCIAL_AUTH_GITLAB_SCOPE to the empty list, as well as allowing it to be set by the settings file or the env.  Of note, the social login adds `read_user` to this list, which is enough for login, and as such the `[]` default should be proper in most cases.

This should fix #434 
This Should fix #431 

I also added a settings dumper that is useful for debugging, and shell scripts, which may help with #432


shub/settings.py
```diff
@@ -668,4 +668,20 @@ def __iter__(self):
     }
 
 if ENABLE_GITLAB_AUTH:  # noqa
-    SOCIAL_AUTH_GITLAB_SCOPE = ["api", "read_user"]
+    if "SOCIAL_AUTH_GITLAB_SCOPE" not in locals():
```
vsoch, line 671: My preference would be to set this as a default empty list, and then check the environment (or other places) to update it, or set.

shub/settings.py
```diff
@@ -668,4 +668,20 @@ def __iter__(self):
...
+        SOCIAL_AUTH_GITLAB_SCOPE=[]
+    SOCIAL_AUTH_GITLAB_SCOPE += get_sregistry_envar_list("SOCIAL_AUTH_GITLAB_SCOPE")
+    SOCIAL_AUTH_GITLAB_SCOPE = list(set(SOCIAL_AUTH_GITLAB_SCOPE))
+
+
+if __name__ == "__main__":
```
vsoch, line 677: Please make this a management script instead, this is a bit klunky.

shub/settings.py
karcaw: I had tried that, but the yaml import code did not pull in the settings when i pre-populated the variable early in the code.

shub/settings.py
karcaw: done!

scripts/show_settings.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python3
+
+import sys
+import shub.settings as cfg
+
+if __name__ == "__main__":
```
vsoch, line 6: This is closer! You'd actually want it to be a proper Django management module - this would allow the user to run like `python manage.py show_settings`. Here is more background https://docs.djangoproject.com/en/4.2/howto/custom-management-commands/ and there is an example [here](https://github.com/singularityhub/sregistry/tree/master/shub/apps/base/management/commands). Basically, you add a script of similar format to that directory, then test running with `manage.py` and add a note to the documentation about running it. So if you had a script called `show_settings.py` in that directory, and it followed the same format as that registry.py example, you could run `python manage.py show_settings.py` let me know if that makes sense!

shub/settings.py
vsoch: I don't follow, sorry. The structure of settings.py is that we have defaults in these structures: https://github.com/singularityhub/sregistry/blob/5477171c8a042344fa2b01be97ba2cb534ba0e91/shub/settings.py#L28 and then we either update that from the settings file here: https://github.com/singularityhub/sregistry/blob/5477171c8a042344fa2b01be97ba2cb534ba0e91/shub/settings.py#L274-L282 - you should be able to follow a similar pattern here.

karcaw: right, but i could find no LIST_DEFAULTS, so i patterned it more how the PLUGINS_ENABLED is done.

vsoch: Maybe make `LIST_DEFAULTS` then, and (if it makes sense) use that for plugins too?

scripts/show_settings.py
karcaw: i'll get django on my hands...  i'll give it a try...

vsoch: Awesome! And if you run into trouble since it's about to be the weekend I can give it a shot too, and at that time feel free to assign me other tasks for stuffs that you are having trouble with. Django is a bunch to learn if you haven't used it before so I'm totally happy to help!

karcaw: think i got this going.  had to pull in the whole dependency list for django it seemed...

shub/settings.py
karcaw: should  the environment overwrite the settings.yaml and the secrets file, or should it append to the list and uniqify it.   currently PLUGINS_LIST appends what is in the env to the list, and i kinda copied that for the GITLAB scope, or should it be a full replacement?

vsoch: The order that we have currently starts with defaults, overrides with environment, and then checks for non None values in the settings.yaml. I would keep that for now, although looking at this again after many years typically you want environment variables to always take precedence. We can shuffle those around in another PR after this one if you agree.

karcaw: i did an overwrite as that matches what other things do.

shub/apps/base/management/commands/show_settings.py
```diff
@@ -0,0 +1,31 @@
+"""
+
+Copyright 2017-2023 Evan Felix.
```
vsoch, line 3: When you contribute to an OSS project you typically have to use the same Copyright headers.

shub/settings.py
```diff
@@ -252,6 +278,11 @@ def get_sregistry_envar_list(key):
         STRING_DEFAULTS[key] = value
 
 
+for key in LIST_DEFAULTS:
+    value = get_sregistry_envar_list(key)
+    if value is not []:
+        LIST_DEFAULTS[key] = list(set(value))
```
vsoch, line 284: We could take two directions here:

- Just set it here if it's not an empty list.
- update the defaults to include it (and take the unique set of all of those)

What do you think?

karcaw: im not sure what you mean with this?  the defaults have the two lists that I wanted, and they get overridden here by env variables.

vsoch: TLDR you can either do:

```python
# totally override, don't care if something was there
if value:
    LIST_DEFAULTS[key] = value
```
or

```python
# be flexible to allowing to extend
if value:
    LIST_DEFAULTS[key] += value
```
And for the latter (if you do that) I'd do a loop over all of them at the end to do a list(set( to ensure uniqueness.

karcaw: make sense, i kinda moved to the overwrite method, as that is how all the other  settings work.

shub/apps/base/management/commands/show_settings.py
karcaw: I just took the header out, as it seems the MPL does not require it.

vsoch:
Looking good! Last item of action - please bump the version in https://github.com/singularityhub/sregistry/blob/master/VERSION and add a note to the CHANGELOG. If you'll be making a few PRs, we will put them all under this new version.

vsoch:
@karcaw for linting, you can install pre-commit locally (and use our requirements.dev.txt to do so) and fix the errors

https://github.com/singularityhub/sregistry/blob/ee23901a00a23d0d2a65fab5f484c803cfd60e44/.github/workflows/main.yml#L21-L29

vsoch (approved):
Nice job @karcaw !

PR #436, opened 2023-06-02
Update scripts and docs to reflect changes in settings file.

karcaw:
This updates a few things that were changed in #409.
  * update scripts to reflect missing config.py, and use new settings.
  * update docs to not refer to old config files, but use the new ones.

Relies on changes in #435

### DIFF
--- a/shub/settings.py
+++ b/shub/settings.py
@@ -668,4 +668,20 @@ if ENABLE_GOOGLE_AUTH:  # noqa
     }
 
 if ENABLE_GITLAB_AUTH:  # noqa
-    SOCIAL_AUTH_GITLAB_SCOPE = ["api", "read_user"]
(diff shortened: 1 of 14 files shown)


vsoch:
This is beautiful - thank you! Let's finish up #409 and get these in too! :tada:

shub/settings.py
```diff
@@ -668,4 +668,20 @@ def __iter__(self):
     }
 
 if ENABLE_GITLAB_AUTH:  # noqa
-    SOCIAL_AUTH_GITLAB_SCOPE = ["api", "read_user"]
+    if "SOCIAL_AUTH_GITLAB_SCOPE" not in locals():
```
vsoch, line 671: Looks like this change from the other PR snuck in!

karcaw: i mentioned that is relies on changes from the #435 because the scripts need that other script to work properly.

